PR #18, opened 2017-03-22
Os eval update

stevenwaslander:
Trying out ditto, and fixing a few spacing issues and image embedding, which can only be done locally, and not remotely.


stevenwaslander:
docs only commits probably shouldn't trigger travis.

chutsu:
ok, so I had a go at it (embedding youtube videos), and I don't think the StackOverflow method is actually needed for our usecase, the complexity of taking a screenshot puting it within the docs folder and then creating a slightly obfulscated link to display the video is a little inconvenient.

I tried using YouTube's own "Share" by "Embed" code and it worked:

```
<div style="position:relative;height:0;padding-bottom:56.25%"><iframe src="https://www.youtube.com/embed/LVV88RZq0YQ?ecver=2" width="640" height="360" frameborder="0" style="position:absolute;width:100%;height:100%;left:0" allowfullscreen></iframe></div>
```

screenshot_2017-03-22_10-46-22

The picture shows the current youtube linking method above, and the embed method below, the second one is better me thinks :)

chutsu:
Second point I would like to bring up is I believe that **all markdown docs should be hard wrapped to 80 column width**, hard wrap meaning there's a new-line character appended at the end of an arbitrary column width (say 80). The rationale for this decision is to aid the peer-review process, if the line was not hardwrapped it becomes very difficult to locate the changes made within the line (think of how we wrote conference papers in the docs repo, it was near impossible to see changes easily). 

Consider the following screenshot example where docs are **not hard-wrapped**:

screenshot_2017-03-22_11-00-57

I know that support for hard-wrapping lines is not supported on every editor, but there's actually a command we can use:

    fold -w 80 -s text.txt

or 

    fmt -w 80 <text.txt

(adding mike to this discussion)

stevenwaslander:
I strongly disagree.  Hard line returns are artificial, and ridiculous and lead to endless formatting headaches when copying text to different editors, and require affecting every line in a paragraph anyway when you insert a single word at the top.  In the picture you posted is is perfectly clear what changed, because github highlights the two changes.

stevenwaslander:
For code, I get it, and don't like 80 characters, but prefer 120.  For text it is totally unhelpful.

stevenwaslander:
Can you make something that hides the iframe from the user.  So a ditto tag that puts all that mess in for us?
e.g. <<vid_link>>

stevenwaslander:
I was ok with the image inside a link formatting, because that is manageable.  But having to look up and paste in a scraggle of html is really annoying.
Already severely missing wsiwyg and simple buttons to do simple tasks.

stevenwaslander:
Usability is essential for docs. Minimum friction.  Code needs to be correct, docs need to be easy.

chutsu:
Ok, I'll have a look at creating a `<<vid>>` like feature just for ditto later in the week.

stevenwaslander:
Thanks.

stevenwaslander:
I'll stop complaining soon, but I worry people will document less if it is more work.  Hopefully, people get used to it quickly.  The PR is a good thing, too, I think now, so two people read docs, and we identify ways to make it all easier.  And I'm guessing people will do chunks of documenting (so large PRs), which should still be easy to review.

stevenwaslander:
By the way, fully agree that the embed looks way better. and not having to crop and save the picture.

michaelhsmart:
Commenting since Chris summoned me.

For line wrapping I partially agree with both of you.

It's wholly artificial by design (single newlines do nothing for markdown). I don't agree that they cause headaches and ridiculousness based on editors, but what I think Steve meant to get at is that they cause headaches and incompatibility if you want to take the content out of markdown. Arbitary example: I copy a paragraph to put into an email. Or a slack message. Or anywhere that doesn't ignore single newlines. Now I have newlines in the middle of sentences. Gross.

That said, github's diff is *terrible* for this kind of diff. Chris just happened to pick the example that shows the opposite of his point (very often those changes will **not** be hightlighted). Recently I tried to figure out the different doc branches on moose (.tex) and it was a colossal mess for the exact reason that git/github's diff's don't resolve real changes within shift changes well.

Often github will just miss the real change within the shift change. This is true for code as well and why I hate formatting changes getting mixed in with content changes. Example (There are actual non-formatting changes in this diff): https://github.com/wavelab/autonomoose/pull/253/files?diff=split#diff-8854d70eecd27735582640b4821b114cL57

While the 80 char limit appears to make it less likely for a real change to be missed by the diff, it throws in a lot of extra crap too when a paragraph adds a single word in the first line and propagates down. I've been using it for all of my own markdown work just because I've found it much easier to deal with for diffs, but I don't think it is a one-size-fits-all solution to github's diff limitations. Especially for the ease-of-use needs for docs.  

Who is really writing enough large, frequently changing documentation items that we need to treat docs like code?

michaelhsmart:
TLDR: I really do like the 80 char limit for the specific types of text files that support it. But this would be the type of thing that I think people can make up their own minds about for their own doc content.

stevenwaslander:
Reviewing docs just makes them more important, and makes more people aware of what went in.  But slows their creation...

chutsu:
its bound to happen, PRs in large companies sometimes takes weeks of review before it is actually merged in.

michaelhsmart:
Ok Last comment on this from me.

The 80 limit objectively does make it easier to view/read for PR. Wrapping severely detracts from readability. Here's an example of what it looks like with wrapping (https://github.com/wavelab/autonomoose/commit/aba91054f94017e2c9e086fd962a2da3424881c3#diff-71a1898ff93063ce750168ae8f70dd36R436)

Note in that example it's an ugly table anyway so there might not be a better solution, but you can get the idea and use your imagination.

My opinion is that the 80 limit is great, but should be discretionary. i.e. don't autoscript it or something insane like that.

stevenwaslander:
Fine on the 80 character thing, although I'll remain a dissenting opinion.  It's definitely worse for prose, and only better for tables and equations and the like if you don't hard wrap those lines yourself for readability.

Issue #19, opened 2017-03-22
Add project-wise C++ compilation warnings


chutsu:
Currently we do not have C++ compilation warnings, this issue addresses that.

PR #20, opened 2017-03-22
Fix 19 cmake warnings

chutsu:
This PR is in relation to #19 

In summary this PR just adds a `-Wall` to `CMAKE_CXX_FLAGS` during compilation


wave_utils/src/config.cpp
```diff
@@ -66,7 +66,7 @@ int ConfigParser::checkVector(std::string key,
     // clang-format on
 
     // check number of values
-    if (this->root[key].size() != vector_size) {
+    if (this->root[key].size() != (size_t) vector_size) {
```
leokoppel: Use C++ cast. Or change the type of `vector_size`.

wave_utils/src/config.cpp
```diff
@@ -122,6 +122,7 @@ int ConfigParser::loadPrimitive(ConfigParam param) {
         case FLOAT: *param.f = node.as<float>(); break;
         case DOUBLE: *param.d = node.as<double>(); break;
         case STRING: *param.s = node.as<std::string>(); break;
+        default: return -3;
```
leokoppel: What is -3? Use enum if it has some meaning.

wave_utils/src/config.cpp
```diff
@@ -142,30 +143,32 @@ int ConfigParser::loadArray(ConfigParam param) {
     this->getYamlNode(param.key, node);
     switch (param.type) {
         case BOOL_ARRAY:
-            for (int i = 0; i < node.size(); i++) {
+            for (size_t i = 0; i < node.size(); i++) {
```
leokoppel: Can use range-based for.

wave_utils/src/config.cpp
chutsu: I'll do another issue to write a doc for this soon. In the mean time let's just keep it this way for the PR? (else this goes beyond the scope of this PR)

wave_utils/src/config.cpp
arunxdas: :+1: static_cast<>

wave_utils/src/config.cpp
chutsu: Done

wave_utils/src/config.cpp
chutsu: Done

PR #24, opened 2017-03-23
Feature 12 rotation kinematics

arunxdas:
This PR implements a subset of the rotation class required by the kinematics library.  The following member functions are implemented in this PR:

```
 fromEulerXYZ();
 fromAngleAxis();
 fromExpMap();
 logMap();
 setToIdentity();
 rotate();
 rotateAndJacobian();
  inverseRotate();
  compose();
  invert();
  isNear();
  manifoldPlus();
  getRotationMatrix();
```

These functions are documented in kinematics.md and rotations.md.  Unit tests are also provided in the tests folder.  The analytical Jacobians are tested through comparison to a central difference method, implemented by the function `numerical_jacobian`.

A simple `skew_symmetric_matrix` function is  added to the math utils library, in order to facilitate some kinematics operations.

CmakeLists for libwave are modified, in order to build the kinematics library.


chutsu:
You need to add a Kindr install script, else travis won't pass.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
+#ifndef __WAVE_NUMERICAL_TEST_FUNCTORS_HPP__
+#define __WAVE_NUMERICAL_TEST_FUNCTORS_HPP__
```
chutsu: No double underscores, its compiler reserved (leo brought this up).

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+#include <kindr/Core>
+
+namespace wave {
+
+// The Functors are used for computing finite difference Jacobians
+// of rotation expressions.
+
+class RotateAndJacobianJpointFunctor {
+public:
+  wave::Rotation R;
```
chutsu: Remove the `wave::`, you are already in the wave namespace

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+  wave::Vec3 P;
+  RotateAndJacobianJpointFunctor(wave::Rotation input_rotation) {
+    this->R = input_rotation;
+  }
+
+  int operator()(wave::Vec3 input_point, wave::Vec3 &output_point) {
+    output_point = R.rotate(input_point);
+  }
+};
+
```
chutsu: Indent spacing does not conform to current 4 space indent, consider running the clang-formatter commands I gave you.

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
+#ifndef __WAVE_KINEMATICS_ROTATION_HPP__
+#define __WAVE_KINEMATICS_ROTATION_HPP__
+
+#include <iostream>
+#include <functional>
+
+#include "wave/utils/utils.hpp"
+#include <kindr/Core>
+
```
chutsu: I tend to folow a header file order of: 

1. C headers
2. C++ headers
3. External library headers
4. Internatl library headers

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
+#include "wave/kinematics/rotation.hpp"
+#include "wave/utils/utils.hpp"
+
+double wave::Rotation::comparision_threshold = 1e-8;
+
+wave::Rotation::Rotation() {}
```
chutsu: Put this whole file in the `wave` namespace?

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
+#include "wave/kinematics/rotation.hpp"
+#include "wave/utils/utils.hpp"
+
+double wave::Rotation::comparision_threshold = 1e-8;
```
chutsu: Put this in header?

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
...
+  double r00 = cz * cy;
+  double r01 = cz * sy * sx - sz * cx;
+  double r02 = cz * sy * cx + sz * sx;
+  double r10 = sz * cy;
+  double r11 = sz * sy * sx + cz * cx;
+  double r12 = sz * sy * cx - cz * sx;
+  double r20 = -sy;
+  double r21 = cy * sx;
+  double r22 = cy * cx;
+
```
chutsu: Should we `const` everything?

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+bool wave::Rotation::isNear(const Rotation &R) {
+  return this->rotation_object.isNear(R.rotation_object,
+                                      this->comparision_threshold);
+}
+
+wave::Mat3 wave::Rotation::getRotationMatrix() {
+  return this->rotation_object.matrix();
+}
```
chutsu: No big deal, but I find this weird, no newline at the end of this curly bracket?

wave_utils/include/wave/utils/math.hpp
```diff
@@ -34,6 +34,7 @@ void vec2mat(std::vector<double> x, int rows, int cols, MatX &y);
 void mat2vec(MatX A, std::vector<double> &x);
 double wrapTo180(double euler_angle);
 double wrapTo360(double euler_angle);
+Mat3 skew_symmetric_matrix(Vec3 a);
```
chutsu: Just a naming thing but do you think if we can, we should borrow names from Matlab? e.g `skewdec` ([matlab doc](http://www.mathworks.com/help/robust/ref/skewdec.html)) for this instanace? I have already borrowed from `wrapTo180` and `wrapTo360`.

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+  wave::numerical_jacobian(
+    Jparam_functor, perturbation_vec, Jparam_numerical);
+  error_matrix = Jparam_analytical - Jparam_numerical;
+  ASSERT_LE(error_matrix.norm(), wave::Rotation::comparision_threshold);
+}
+
+int main(int argc, char *argv[]) {
+  testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
```
chutsu: At current I have a general `wave_utils_tests` and `wave_optimization_tests` test runner for each module of libwave, maybe you should do the same? For example this is how I link up all my test cases into a single test runner in the `utils` module:

```
# UNIT TESTS
SET(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/tests)
FILE(COPY tests/data DESTINATION ${PROJECT_BINARY_DIR}/tests)

ADD_EXECUTABLE(
    # TARGET
    wave_utils_tests
    # TESTS
    tests/utils/config_test.cpp
    tests/utils/data_test.cpp
    tests/utils/filesystem_test.cpp
    tests/utils/math_test.cpp
    tests/utils/time_test.cpp
    # TEST RUNNER
    tests/test_runner.cpp
)
TARGET_LINK_LIBRARIES(wave_utils_tests ${TEST_DEPS})
```

Open to suggestions if you have alternatives.

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+  wave::Mat3 Jparam_analytical, Jpoint_analytical;
+  rotation1.rotateAndJacobian(v1, Jpoint_analytical, Jparam_analytical);
+
+  // Now get the numerical Jacobians
+  wave::Mat3 Jparam_numerical(3, 3), Jpoint_numerical(3, 3);
+
+  // Jacobian wrt the point.
+  wave::RotateAndJacobianJpointFunctor Jpoint_functor(rotation1);
+  wave::numerical_jacobian(Jpoint_functor, v1, Jpoint_numerical);
+  Eigen::MatrixXd error_matrix = Jpoint_analytical - Jpoint_numerical;
```
chutsu: `Eigen::MatrixXd` -> `MatX`?

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+  wave::Vec3 x = _x;
+  
+  for (int j = 0; j < n; ++j) {
+    h = eps * abs(x[j]);
+    if (h == 0.) {
+      h = eps;
+    }
+
+    x[j] += h;
+    F.operator()(x, val2);
```
chutsu: Should this not be `F(x, val2)` since you've implemented the bracket operator?

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+                        Eigen::MatrixBase<MatrixType> &jac) {
+  using std::sqrt;
+  double h;
+  int n = _x.size();
+  const double eps = sqrt(1e-10);
+  wave::Vec3 val1, val2;
+  wave::Vec3 x = _x;
+  
+  for (int j = 0; j < n; ++j) {
+    h = eps * abs(x[j]);
```
chutsu: `abs` returns `int` is this intentional? for `double` return use `fabs()`

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+// Implement the numerical differentiation.  Based on the Eigen::NumericalDiff
+// Module.
+template <typename MatrixType, typename FunctorType>
+void numerical_jacobian(FunctorType F,
+                        wave::Vec3 _x,
+                        Eigen::MatrixBase<MatrixType> &jac) {
+  using std::sqrt;
+  double h;
+  int n = _x.size();
+  const double eps = sqrt(1e-10);
```
chutsu: magic number! you could put this as part of the optional argument 

```
void numerical_jacobian(FunctorType F, 
                                       wave::Vec3 _x, 
                                       Eigen::MatrixBase<MatrixType> &jac, 
                                       const double eps = <some default value>)
```

Or alternatives, but I don't think having a magic number here is a good idea

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+// Implement the numerical differentiation.  Based on the Eigen::NumericalDiff
+// Module.
+template <typename MatrixType, typename FunctorType>
+void numerical_jacobian(FunctorType F,
+                        wave::Vec3 _x,
+                        Eigen::MatrixBase<MatrixType> &jac) {
+  using std::sqrt;
+  double h;
+  int n = _x.size();
```
chutsu: `_x`? not sure we want to use pre-pending underscore variable names.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+  const double eps = sqrt(1e-10);
+  wave::Vec3 val1, val2;
+  wave::Vec3 x = _x;
+  
+  for (int j = 0; j < n; ++j) {
+    h = eps * abs(x[j]);
+    if (h == 0.) {
```
chutsu: Floating point comparisons are hard in C++, need a better way to do this than just an `==`, consider something like my `fltcmp()` in `utils` its by no means perfect, but its a better solution.

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
...
+void wave::Rotation::setToIdentity() {
+  this->rotation_object.setIdentity();
+}
+
+void wave::Rotation::compose(const Rotation &R) {
+  this->rotation_object = this->rotation_object * R.rotation_object;
+}
+
+wave::Mat3 wave::Rotation::invert() {
+  this->rotation_object.invert();
```
chutsu: missing return?

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+  // A vector v1.
+  wave::Vec3 v1;
+  // v1 rotated by rotation1.
+  wave::Vec3 v1_rotated;
+  // True angle axis params for rotation1.
+  wave::Vec3 rotation1_angleaxis_params;
+
+  void SetUp() {
+    wave::Rotation::fromEulerXYZ(0.1, 0.2, 0.3, rotation1);
```
chutsu: `this->rotation1`

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+      0.288748939228675;
+  }
+};
+
+
+// Construction tests.
+
+// Test that fromEulerXYZ returns the correct matrix.
+TEST_F(RotationTestFixture, test_from_euler_xyz) {
+  ASSERT_TRUE(rotation1_truth.isNear(rotation1));
```
chutsu: ASSERT_TRUE(this->rotation1_truth.isNear(this->rotation1));

wave_utils/src/math.cpp
```diff
@@ -86,4 +86,12 @@ double wrapTo360(double euler_angle) {
     }
 }
 
+Mat3 skew_symmetric_matrix(Vec3 a){
+  Mat3 skew_matrix;
+  skew_matrix << 0, -a(2), a(1),
```
chutsu: If you're using clang-format it will put the following in 1 line, consider doing this:

```
  Mat3 skew_matrix;
  // clang-format off
  skew_matrix << 0, -a(2), a(1),
                a(2), 0, -a(0),
                -a(1), a(0), 0;
  // clang-format on
```

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
chutsu: Or even have it as a class member variable, open to suggestions

wave_kinematics/src/rotation.cpp
leokoppel: It's an initialization of a non-const static so you can't.. But it should not be a non-const static.

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+
+// Test to ensure setToIdentity returns identity.
+
+TEST_F(RotationTestFixture, test_set_to_identity) {
+  // First make sure the rotation is not already identity.
+  ASSERT_FALSE(rotation1.isNear(eye3));
+  rotation1.setToIdentity();
+  ASSERT_TRUE(rotation1.isNear(eye3));
```
leokoppel: For my tests I use a helper predicate like this: http://stackoverflow.com/a/39238772/431033

The benefit is if it fails, gtest will print both matrices.

(Except that probably won't work here, because the Rotation type has no printing method defined - another reason not to wrap it)

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+// Test functionality of the boxplus operator.
+TEST_F(RotationTestFixture, test_manifold_plus) {
+  // We know that rotation3 = rotation1*rotation2
+  // Therefore rotation3 = exp(rotation1_angleaxis_params)*rotation2,
+  // which is exactly the boxplus operation.
+
+  // Make sure rotation2 and rotation3 are not the same initially
+  ASSERT_FALSE(rotation2.isNear(rotation3));
```
leokoppel: See above. This is a bit confusing. You should not test the test fixture.

wave_kinematics/tests/rotation_test.cpp
leokoppel: Most of the stuff in `RotationTestFixture` would be better placed in the individual tests which check those functions.  A test fixture makes sense if you have to generate a bunch of input data that comes from _not_ the functions under test. A single result _from_ the FUT does not belong in the fixture.

See http://blog.stevensanderson.com/2009/08/24/writing-great-unit-tests-best-and-worst-practises/ 
> Avoid having common setup code that runs at the beginning of lots of unrelated tests. Otherwise, it’s unclear what assumptions each test relies on, and indicates that you’re not testing just a single unit.

scripts/install/kindr_install.bash
```diff
@@ -0,0 +1,28 @@
...
+}
+
+install_kindr() {
+    # compile and install kindr
+    cd /usr/local/src/kindr
+    cd kindr-$KINDR_VER
+    mkdir -p build
+    cd build
+    cmake ..
+    sudo make install
```
leokoppel: We robably want some -j option right? @chutsu 

Also general comment, `make install` is the only command that should use sudo. Everything else, including `make`, ideally should not.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
chutsu: oh dear, this is potentially dangerous the `abs` and `std::abs` are quite different (`abs` returns integer, `std::abs` returns float, double or long double depending on input type)

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
chutsu: Not quite true, in that Eigen example `h` is of type `Scalar` (I can't find docs for it, atm). our `h` is of type `double` so its still a concern. Maybe it would be better to write this?

```
if (h < 0.0) {
    h = eps; // make sure h is non-zero
}
```

If that is the purpose.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
leokoppel: The dangerous part is that sometimes `::abs` is double. "The problem is that calling the overloaded abs(x) only works as intended if the right header has been included. " - https://developers.redhat.com/blog/2016/02/29/why-cstdlib-is-more-complicated-than-you-might-think/

wave_kinematics/tests/rotation_test.cpp
chutsu: How so? From a coding standards point view Me and Mike are already in agreement the `this` specifies where the variable came from.

leokoppel: ..yeah I take it back. (I don't like `this->` in general but a unit test is no different). 

Really to improve legibility, neither of those variables should be part of the test fixture.

wave_utils/src/math.cpp
chutsu: I don't really see it as a big problem actually, because:
- the directive is short and well defined 
- I intend to ask people to use the clang-format tool to help speed up formatting (common format without question)
- clang-format is IDE-agnostic
- removing clang-format directives is literally 2 commands away 
```
    cat file.cpp | '/\/\/\ clang\-format\ off/,+1 d'
    cat file.cpp | '/\/\/\ clang\-format\ on/,+1 d'
```

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
leokoppel: `Scalar` is just a typename. When you supply a functor to Eigen, it uses your scalar type. For example, `double`. 

As I mentioned in my other comment, the default `eps` is machine epsilon. If the algorithm works when `h` is machine epsilon, it will work when `h` is _n_*machine epsilon. Unless of course it's exactly zero.

scripts/install/kindr_install.bash
chutsu: - `-j` option may be in future, but for a simple build we are really pre-mature optimizing here, this literally takes 3 seconds to build with default settings.
- I would agree, but in this case almost all the comands run in `/usr/local/` which requires sudo access, unless I'm wrong?

leokoppel: Oh yeah, 3 seconds, never mind.

You are right. We have already talked about the choice of directory : )

wave_kinematics/src/rotation.cpp
arunxdas: Kindr's Euler methods were not giving the expected output.  I looked into it briefly, and it looks like they have these automatic conversion methods which are not easily decipherable unless you take a really detailed look.  As a quick work around, I coded this up because we've used it before in other projects and know it's correct, but I'll try Eigen's method.

wave_utils/include/wave/utils/math.hpp
arunxdas: I'd prefer naming that is easily understood first time around, without having to have familiarity with matlab's convention.  I find skewdec a bit unclear, where as wrapTo180 and wrapTo360 seem pretty clear already.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: Yes, there is a floating point abs, but I can switch to fabs for clarity.

wave_kinematics/tests/rotation_test.cpp
arunxdas: The variables are used in all tests, though.  Should they be removed from test fixtures and just duplicated for each test?

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace wave {
+
+class Rotation {
+public:
+  // Internal storage as a kindr Rotation object.
+  kindr::RotationMatrixD rotation_object;
```
arunxdas: The goal of the wrapper is to provide an interface that is independent of the underlying implementation.  If we one day decide to dump Kindr and move to Sophis, TooN, or even just Eigen, the change is encapsulated within the kinematics classes.  To the rest of our code, the change is transparent.  Is this still possible to do if we inherit from RotationMatrix, or would we be stuck with their member function names and signatures?

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
```diff
@@ -0,0 +1,78 @@
...
+  int operator()(wave::Vec3 input_point, wave::Vec3 &output_point) {
+    wave::Rotation Rp = this->R;
+    Rp.manifoldPlus(input_point);
+    output_point = Rp.rotate(P);
+  }
+};
+
+
+// Implement the numerical differentiation.  Based on the Eigen::NumericalDiff
+// Module.
```
arunxdas: I'll be implementing more member functions after this PR, which are expressions that return Rotation Objects.  To find the numerical Jacobians of such expressions, we need to replace normal minus with manifoldMinus when computing (val2 - val1), thus the numerical_jacobian function will undergo quite a few more changes.  Are licensing issues still a concern, in that case?

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
...
+double wave::Rotation::comparision_threshold = 1e-8;
+
+wave::Rotation::Rotation() {}
+
+void wave::Rotation::setToIdentity() {
+  this->rotation_object.setIdentity();
+}
+
+void wave::Rotation::compose(const Rotation &R) {
+  this->rotation_object = this->rotation_object * R.rotation_object;
```
arunxdas: I'll implement an overloaded * operator, but those are mainly for convenience.  The point of having a dedicated compose expression is that there will also be a matching composeAndJacobian, which is much stricter on defining a **right** sided composition, and therefore Jacobians of the expression with respect to the left rotation, and Jacobians wrt the right rotation.  The big picture here, is that the kinematics library supplies Jacobians of these elementary operations, such that when we go to implement our residual functions, the Jacobian of the residual is easily calculated through chain rule of the elementary functions.  

I believe we should get into the habit of viewing these residual expressions as a series of composed functions, which I think is easier to do when explicitly in function form, rather than using operators.

wave_utils/src/math.cpp
leokoppel: Yeah clang-format is good, just this one little shortcoming. [This method](http://stackoverflow.com/a/34362422/431033) seems OK for a small matrix (`//` at the end of each line).

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
...
+  double r11 = sz * sy * sx + cz * cx;
+  double r12 = sz * sy * cx - cz * sx;
+  double r20 = -sy;
+  double r21 = cy * sx;
+  double r22 = cy * cx;
+
+
+  R.rotation_object.setMatrix(r00, r01, r02, r10, r11, r12, r20, r21, r22);
+
+  return true;
```
arunxdas: see previous comment regarding kindr's EulerAnglesXyz

wave_kinematics/tests/rotation_test.cpp
arunxdas: ahh ok, so this answers my previous question.  So then the same variables are to be duplicated in the individual tests?

wave_kinematics/tests/rotation_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+    rotation1_angleaxis_params << 0.068924613882066, 0.213225926957886,
+      0.288748939228675;
+  }
+};
+
+
+// Construction tests.
+
+// Test that fromEulerXYZ returns the correct matrix.
+TEST_F(RotationTestFixture, test_from_euler_xyz) {
```
chutsu: Having another look at the Gtest doc, it says: "Both names must be valid C++ identifiers, and they should not contain underscore (_).", this is not a "must not" but rather a "should", that said I believe we should follow the guidelines as much as possible. ([source](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#simple-tests))

chutsu: A slightly more detailed explanation can be found [here](https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-should-not-test-case-names-and-test-names-contain-underscore).

wave_kinematics/src/rotation.cpp
leokoppel: Oh, it looks like `fromEulerXYZ(a, b, c, ...)` corresponds to `kindr::EulerAnglesZyxD(c, b, a)`. See  the rotation matrix in [Kindr's unit test](https://github.com/ethz-asl/kindr/blob/master/test/rotations/EulerAnglesZyxTest.cpp#L718).

I.e. kindr's notation `EulerAnglesXyz` confusingly means something different than your `EulerXYZ`. Maybe you could add some clarifying words such as "body-fixed" or "intrinsic" or "extrinsic"  (whatever is correct) to the docs?

By the way if you change the function, the unit test can compare against the explicit "sin() + cos()" version. I think that is clearer than a pre-calculated "truth" `0.936293363584199...` of unclear origin.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
leokoppel: Yes. Eigen's MPL license is copyleft at file level. It looks like any files containing modified MPL-licensed code must be distributed under MPL. See [MPL 2.0 FAQ](https://www.mozilla.org/en-US/MPL/2.0/FAQ/)

A few possibilities:

* Put Eigen-derived code in a separate file and adhere to license
* Wipe your memory and implement the algorithm separately
* Use Eigen's `NumericalDiff.h` and define an `operator-` for your value type. It will be used.

wave_kinematics/include/wave/kinematics/rotation.hpp
leokoppel: That's a tough question. Yes, if you want different function names and signatures you will have to wrap everything.

Right now, the difficulty of designing a good interface seems much greater than the difficulty of changing client calls. If there were 10 third party libraries depending on the `Rotation` class that would be different, but I don't see how we'd get to that point without a really solid, versatile interface. And the law of leaky abstractions suggests that we would have to change some client code, anyway.

For example, Kindr itself wraps Eigen objects but it takes thousands of lines of code to do so.

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+                         Mat3 &Jpoint,
+                         Mat3 &Jparam);
+  Vec3 inverseRotate(const Vec3 &input_vector);
+  void compose(const Rotation &R);
+  Mat3 invert();
+  bool isNear(const Rotation &R);
+  void manifoldPlus(const Vec3 omega);
+
+  // Set and get methods.
+  Mat3 getRotationMatrix();
```
leokoppel: Be const correct.
`to` seems more appropriate than `get` as it is a conversion (also matches Eigen).

```cpp
Mat3 toRotationMatrix() const;
```

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+  // Module functions.
+  Vec3 logMap();
+  void setToIdentity();
+  Vec3 rotate(const Vec3 &input_vector);
+  Vec3 rotateAndJacobian(const Vec3 &input_vector,
+                         Mat3 &Jpoint,
+                         Mat3 &Jparam);
+  Vec3 inverseRotate(const Vec3 &input_vector);
+  void compose(const Rotation &R);
```
leokoppel: This prototype makes sense; clear what it does.

It may make sense to add another
```
Rotation composed(const Rotation&) const;
```

With this particular operation though, why not stick to * operator?

wave_kinematics/src/rotation.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+wave::Vec3 wave::Rotation::rotate(const Vec3 &input_vector) {
+  return this->rotation_object.rotate(input_vector);
+}
+
+wave::Vec3 wave::Rotation::inverseRotate(const Vec3 &input_vector) {
+  return this->rotation_object.inverseRotate(input_vector);
+}
+
+wave::Vec3 wave::Rotation::rotateAndJacobian(const Vec3 &input_vector,
```
leokoppel: This method seems to do three things which could be separated.

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+                           const double rotation_y,
+                           const double rotation_z,
+                           Rotation &R);
+  static bool fromAngleAxis(const double angle_magnitude,
+                            const Vec3 rotation_axis,
+                            Rotation &R);
+  static bool fromExpMap(const Vec3 se3_vector, Rotation &R);
+
+  // Module functions.
+  Vec3 logMap();
```
leokoppel:
```cpp
Vec3 logMap() const;
```

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+  void setToIdentity();
+  Vec3 rotate(const Vec3 &input_vector);
```
leokoppel:
```cpp
const
```

same for any method which does not touch your object (`rotateAndJacobian`, `inverseRotate`, `isNear`)

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+  Vec3 rotateAndJacobian(const Vec3 &input_vector,
+                         Mat3 &Jpoint,
+                         Mat3 &Jparam);
+  Vec3 inverseRotate(const Vec3 &input_vector);
+  void compose(const Rotation &R);
+  Mat3 invert();
+  bool isNear(const Rotation &R);
+  void manifoldPlus(const Vec3 omega);
```
leokoppel: `const Vec3&` for argument
It's clear what this does, but may be more convenient to define
```cpp
Rotation manifoldPlus(const Vec3&) const;
```

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+  Vec3 logMap();
+  void setToIdentity();
+  Vec3 rotate(const Vec3 &input_vector);
+  Vec3 rotateAndJacobian(const Vec3 &input_vector,
+                         Mat3 &Jpoint,
+                         Mat3 &Jparam);
+  Vec3 inverseRotate(const Vec3 &input_vector);
+  void compose(const Rotation &R);
+  Mat3 invert();
+  bool isNear(const Rotation &R);
```
leokoppel: Take tolerance argument (as already noted)

docs/api/kinematics/rotation.md
```diff
@@ -0,0 +1,189 @@
...
+
+The threshold that is used to compare if rotation matrices are sufficiently close to one another.  see isNear() function for more details.
+```
+static double comparision_threshold;
+```
+- - - -
+
+**Class Member Functions**:
+
+## Rotation()
```
leokoppel: Hey @chutsu, have you considered documentation from doxygen, etc? Now if Arun changes the interface he will have to manually go through and change everything here even when it's redundant.

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+  // "close enough" to each other.  Used in unit testing and the "isNear"
+  // function.
+  static double comparision_threshold;
+
+public:
+  EIGEN_MAKE_ALIGNED_OPERATOR_NEW
+
+  // Rotation Object Initializers.
+  Rotation();
+  static bool fromEulerXYZ(const double rotation_x,
```
arunxdas: So with these prototypes, does it change the object in-place, and also return a ref to that object? or a ref to a new object, but doesn't touch "this" object?

wave_kinematics/include/wave/kinematics/rotation.hpp
arunxdas: Rotation and not a ref to a rotation?

wave_kinematics/include/wave/kinematics/rotation.hpp
leokoppel: It does touch the object. You can tell because there is no `const` on the end.

Returning a reference is a common pattern that allows chaining, and is used in Eigen. For example 
`r.setFromEulerXyz(x).invert()`.

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: done

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: done

wave_kinematics/include/wave/kinematics/rotation.hpp
arunxdas: done

wave_kinematics/src/rotation.cpp
arunxdas: done

wave_kinematics/src/rotation.cpp
arunxdas: changed comparison_threshold to be argument of isNear

wave_kinematics/src/rotation.cpp
arunxdas: added newline at end of file

wave_utils/include/wave/utils/math.hpp
arunxdas: changed to kindr's getSkewMatrixFromVector

wave_kinematics/tests/rotation_test.cpp
arunxdas: done

wave_utils/src/math.cpp
arunxdas: function removed, changed to kindr's ssm

wave_kinematics/include/wave/kinematics/rotation.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace wave {
+
+class Rotation {
+public:
+  // Internal storage as a kindr Rotation object.
+  kindr::RotationMatrixD rotation_object;
+  // The comparision threshold used to test if rotation objects are
+  // "close enough" to each other.  Used in unit testing and the "isNear"
+  // function.
+  static double comparision_threshold;
```
arunxdas: added comparison threshold as an argument to isNear()

wave_utils/include/wave/utils/math.hpp
```diff
@@ -34,6 +34,7 @@ void vec2mat(std::vector<double> x, int rows, int cols, MatX &y);
...
```
arunxdas: done

wave_kinematics/include/wave/kinematics/rotation.hpp
arunxdas: what is the difference between "compose" and "composed" ? Yes, will  do * operator for just compose

wave_kinematics/include/wave/kinematics/rotation.hpp
arunxdas: done

wave_kinematics/include/wave/kinematics/rotation.hpp
leokoppel:
```
Rotation r1, r2;
...
r1.compose(r2); // modifies r1
auto r3 = r1.composed(r2); // does not modify r1
```
The equivalent is `*=` and `*`.

wave_kinematics/include/wave/kinematics/rotation.hpp
arunxdas: Changed interface as per suggestions

wave_kinematics/src/rotation.cpp
arunxdas: used member functions instead of underlying implementation to make these operations more clear

leokoppel: Not sure I understand. Is "used member functions " a new change to the code in this PR, or are you explaining your original choice?

To me, calls of the form
```cpp
auto rotated_vector = someRotateFunction(input_vector);
auto jacobian = someJacobian(input_vector);
auto other_jacobian = anotherJacobian(input_vector);
```
are more clear than
```cpp
Mat3 jacobian; // uninitialized
Mat3 other_jacobian; // uninitialized
 // Note: can't tell which parameter is which without looking it up
Vec3 rotated_vector = rotateAndJacobian(input_vector, jacobian, other_jacobian);
```

arunxdas: These are new code changes that I'll push in this PR, just responding to the comments.

In this case someJacobian, and anotherJacobian are trivial calculations, so I don't think they warrant their own functions.

leokoppel: There's little cost to having multiple functions (unless the calculations are _expensive_, but cheaper when combined). To the contrary, combining operations increases complexity for the user.

arunxdas: ahh ok, so the suggestion is to decouple the function and jacobian.  I disagree that is clearer for this use case.  The Jacobian is always specific to the input and output relationship to the function.  For example, if there is a function F that takes in 5 different parameters (not necessarily vector elements, could be manifold quantities):
``` y = F(x1,x2,x3,x4,x5) ```
I think it's cleaner to have the function
``` FandJacobian(x1,x2,x3,x4,x5, J_x1, J_x2, J_x3 ...)```
Which is pretty clear in saying that J_x1 is the jacobian of the underlying function F, wrt the x1 parameter, etc, rather than
```  y = F(x1,x2,x3,x4,x5 ) 
    dy_x1 = FandJacobianX1(x1,x2,x3,x4,x5)
    dy_x2 = FandJacobianX2(x1,x2,x3,x4,x5)
  ...
```
Which feels like unnecessary bloat to me, and sometimes will be duplicating calculations since we will loose intermediate calculations

leokoppel: OK, that makes sense. I guess it is similar to a ceres cost function, which takes arrays of _n_ parameter blocks, _n_ outputs, and _n_ Jacobians.

My clarity concern came from the function in the PR having one input and two Jacobians. Yes, if you have _n_ inputs and _n_ Jacobians, it is clear which one is which.

wave_kinematics/tests/rotation_test.cpp
arunxdas: Changed to use your test runner

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: done, along with re-write numerical diff function

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: fixed

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: yes, just checking for exactly zero. I added comments to make this clear

wave_kinematics/src/rotation.cpp
arunxdas: fixed

wave_kinematics/include/wave/kinematics/numerical_test_functors.hpp
arunxdas: re-wrote numerical diff function

arunxdas:
OK to merge?

chutsu:
Looks good to me, Square and merge it, and edit the commit summary so it only has core changes.

Issue #7, opened 2017-03-13
PCL Wrapper


chutsu:

Jebediah:
Overall issue for

- [ ] #11 
- [ ] #8 
- [ ] #9

PR #28, opened 2017-03-27
Feature 11 pcl icp

Jebediah:
This PR is add a wrapper for PCL's ICP matcher, see #11 and #7. The covariance estimation is to come in a future PR, the focus for this one is the general matcher interface, should speed up PR's for other matchers

Summary:

- Add base class to derive specific matchers from
- Add ICP class that wraps PCL ICP
- Add PCL to the install scripts
- Add a few tests to try interface


arunxdas:
PRs must also include api documentation

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
+#ifndef __WAVE_MATCHING_MATCHER_HPP__
+#define __WAVE_MATCHING_MATCHER_HPP__
+
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+
+namespace wave {
+namespace matching {
```
chutsu: I'm not sure we need the matching name space, I think we would like to avoid having additional namespaces if we can.

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+class ICPTest : public testing::Test {
+  protected:
+    ICPTest() {}
+    virtual ~ICPTest() {}
+    virtual void SetUp() {
+        ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        pcl::io::loadPCDFile("../../wave_matching/tests/testscan.pcd", *ref);
+    }
+    void setparams(const float res, const Eigen::Affine3d perturb) {
```
chutsu: Camel Case

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <gtest/gtest.h>
+#include <pcl/io/pcd_io.h>
+#include <Eigen/Dense>
+#include <iostream>
+
+#include "wave/matching/ICP.hpp"
+
+bool isApprox(const Eigen::Affine3d& first,
+              const Eigen::Affine3d& second,
+              double TOL) {
```
chutsu: tolerance? why all upper case?

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+  protected:
+    ICPTest() {}
+    virtual ~ICPTest() {}
+    virtual void SetUp() {
+        ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        pcl::io::loadPCDFile("../../wave_matching/tests/testscan.pcd", *ref);
+    }
+    void setparams(const float res, const Eigen::Affine3d perturb) {
+        matcher = new wave::matching::ICP_Matcher(res);
```
chutsu: why `new`? where is delete?

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
+// Tests for wrapper
+
+#include <gtest/gtest.h>
+#include <pcl/io/pcd_io.h>
+#include <Eigen/Dense>
+#include <iostream>
+
+#include "wave/matching/ICP.hpp"
+
```
chutsu: Consider putting this whole test file into the wave namespace

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+    void setparams(const float res, const Eigen::Affine3d perturb) {
+        matcher = new wave::matching::ICP_Matcher(res);
+        pcl::transformPointCloud(*ref, *target, perturb);
+        matcher->setup(ref, target);
+    }
+    pcl::PointCloud<pcl::PointXYZ>::Ptr ref, target;
+    wave::matching::ICP_Matcher *matcher;
+};
+
+TEST(icp_tests, initialization) {
```
chutsu: `icp_tests`? You mean `ICPTests`

wave_matching/src/ICP.cpp
```diff
@@ -0,0 +1,57 @@
...
+        this->filter.setInputCloud(ref);
+        this->filter.filter(*(this->ref));
+    } else {
+        this->ref = ref;
+    }
+    this->icp.setInputSource(this->ref);
+}
+
+void ICP_Matcher::setTarget(const PCLPointCloud &target) {
+    if(resolution > 0) {
```
chutsu: `this->`

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+    virtual void setRef(const R& ref) = 0;
+    virtual void setTarget(const R& target) = 0;
+    void setup(const R& ref, const R& target) {
+        this->setRef(ref);
+        this->setTarget(target);
+    };
+    // intended to actually perform the matching, and
+    // then populate the private variables.
+    // should return true if successful
+    virtual bool match() {return 0;}
```
chutsu: Typically the method declaration should follow a logical order, with:

1. Constructors
2. Useful methods
3. Setters and Getters

you currently have 
1. Generic acccessors
2. Constructors
3. Generic accessors
4. Useful Methods

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+  protected:
+    // internal storage
```
chutsu: Superflous comment

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Small displacement using voxel downsampling
+TEST_F(ICPTest, smallDisplacement) {
+    Eigen::Affine3d perturb = Eigen::Affine3d::Identity();
+    perturb.translation() << 0.2, 0, 0;
+    setparams(0.05f, perturb);
+    EXPECT_TRUE(matcher->match());
+    Eigen::Affine3d result = matcher->get_result();
+    EXPECT_TRUE(isApprox(result, perturb, 0.1));
```
chutsu: everything in your tests are a little muddle up and hard to read, I personally would have done it this way:

```
Eigen::Affine3d perturb;
Eigen::Affine3d result 

// setup
perturb = Eigen::Affine3d::Identity();
perturb.translation() << 0.2, 0, 0;
setparams(0.05f, perturb);

// test and assert
result = matcher->get_result();
EXPECT_TRUE(matcher->match());
EXPECT_TRUE(isApprox(result, perturb, 0.1));
```

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+bool isApprox(const Eigen::Affine3d& first,
+              const Eigen::Affine3d& second,
+              double TOL) {
+    for (uint8_t iii = 0; iii < 4; ++iii) {
+        for(uint8_t jjj = 0; jjj < 4; ++jjj) {
+            TOL -= std::abs(first(iii,jjj) - second(iii,jjj));
+        }
+    }
+    if (TOL > 0) {
```
chutsu: Floating point comparisons, this may never be true. http://floating-point-gui.de/errors/comparison/

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <pcl/io/pcd_io.h>
+#include <Eigen/Dense>
+#include <iostream>
+
+#include "wave/matching/ICP.hpp"
+
+bool isApprox(const Eigen::Affine3d& first,
+              const Eigen::Affine3d& second,
+              double TOL) {
+    for (uint8_t iii = 0; iii < 4; ++iii) {
```
chutsu: Why `uint8_t` why not normal int?

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return false;
+}
+
+// Fixture to load same pointcloud all the time
+class ICPTest : public testing::Test {
+  protected:
+    ICPTest() {}
+    virtual ~ICPTest() {}
+    virtual void SetUp() {
+        ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
```
chutsu: `this->`

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+        target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
```
chutsu: `this->`

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
+// Tests for wrapper
+
+#include <gtest/gtest.h>
+#include <pcl/io/pcd_io.h>
+#include <Eigen/Dense>
+#include <iostream>
```
chutsu: header file ordering is:

1. C 
2. C++ 
3. Third Party 
4. Internal

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace wave {
+namespace matching {
+
+// sensor data type
+template <typename R>
+class Matcher {
+  public:
+    // generic accessors
+    const Eigen::Affine3d& get_result() {return result;};
```
chutsu: - `this->result`
- `getResult()`

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+    const Eigen::MatrixXd& get_info() {return information;};
```
chutsu: - `this->information`
- `getInfo()`

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+    const float& getRes() {return resolution;};
```
chutsu: - Why a reference `&` to the float? just return float?
- `this->resolution`

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+#define __WAVE_MATCHING_MATCHER_HPP__
+
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+
+namespace wave {
+namespace matching {
+
+// sensor data type
+template <typename R>
```
chutsu: Would have used `T`instead of `R` for typename, unless you have a good reason to choose `R`?

wave_matching/tests/icp_tests.cpp
michaelhsmart: camelCase

chutsu: actually both camelCase and CamelCase are both camel case.

wave_matching/tests/icp_tests.cpp
Jebediah: But it's not checking if it is equal to 0, just greater than. Shouldn't that work even with the floating point error?

wave_matching/include/wave/matching/matcher.hpp
Jebediah: why not?

wave_matching/tests/icp_tests.cpp
Jebediah: why?

wave_matching/tests/icp_tests.cpp
michaelhsmart: Full pedant: lowerCamelCase vs UpperCamelCase

:)

wave_matching/tests/icp_tests.cpp
Jebediah: It's only going to 3, seemed like a waste of an int

chutsu: pre-mature optimization is the root of all evil.

wave_matching/tests/icp_tests.cpp
chutsu: ah touche, was too quick on this one.

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include "wave/matching/ICP.hpp"
+
+bool isApprox(const Eigen::Affine3d& first,
+              const Eigen::Affine3d& second,
+              double TOL) {
+    for (uint8_t iii = 0; iii < 4; ++iii) {
+        for(uint8_t jjj = 0; jjj < 4; ++jjj) {
+            TOL -= std::abs(first(iii,jjj) - second(iii,jjj));
```
michaelhsmart: -= on a pass by value argument?

Does that not trigger a compiler warning?

Jebediah: Should it? It is a copy so it's not going to hurt the original.

stevenwaslander: Can we not use iii and jjj in for loops?

michaelhsmart: I would have assumed it would. I can't imagine intended cases of that are as common as bug cases.

michaelhsmart: This was an honest question, btw. Not a review.

wave_matching/tests/icp_tests.cpp
Jebediah: please

wave_matching/include/wave/matching/matcher.hpp
chutsu: because:
1. float is trivially small, pass by reference is good when you're passing large objects
2. you're returning a reference of `resolution` rather than value, but I'm not actually certain if `resolution` changes, it would also change the reference, e.g.

```
// suppose I do
float r;
r = Matcher.getRes();
// Matcher.resolution changes
r; // did r change as well? I'm not certain r will change as well, so why not just return a value instead of a reference
```

wave_matching/tests/icp_tests.cpp
stevenwaslander: So, what is going on here? Can someone just suggest the correction.  setParams, right?

wave_matching/include/wave/matching/matcher.hpp
stevenwaslander: I would suggest to keep things simple.  Don't unecessarily pass by reference such a tiny thing.  Another optimization that is not warranted.

wave_matching/tests/icp_tests.cpp
chutsu: To save you from having to prepend `wave::` infront of your wave functions / classes. Plus all tests will be using the namespace `wave` so to be consistent I would advise you do the same even though you only use it once here.

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,43 @@
...
+class Matcher {
+  public:
```
stevenwaslander: Should be part of the checklist, used clang? check.  Was there a checklist?

wave_matching/tests/icp_tests.cpp
chutsu: w.r.t `TOL -= ..` its fine in this case.

scripts/install/pcl_install.bash
```diff
@@ -0,0 +1,19 @@
...
+        libvtk6-dev \
+        libeigen3-dev \
+        libusb-1.0-0-dev
+}
+
+echo "Installing PCL ..."
+
+install_dependencies
+
+sudo add-apt-repository -y ppa:v-launchpad-jochen-sprickerhof-de/pcl
```
stevenwaslander: Do we have to use a PPA version? should we be compiling from source instead or using the one in ROS Kinetic?

.gitignore
```diff
@@ -21,3 +21,7 @@ build
 # matlab / octave
 *.mat
 octave-workspace
+
+# clion
+.idea/
+cmake-build-*/
```
stevenwaslander: Are these changes required for this PR? Are they related? I expect we want a universal .gitignore that includes them, but we want to avoid changing them to whatever people want to use locally on random PRs. Don't we have a standard for .gitignore?

wave_matching/tests/icp_tests.cpp
chutsu: 1 last thing, admittedly stupid detail its up to you, but consider using `i++` instead of `++i`. While both are perfectly valid I rarely (if ever) see `++i`. 

Once you're outside the norm, subtle off by one errors creep in. For example these two loops do different things.

```
int i = 0;
int j = i;
while(j < 5) {
    printf("%d", i);
    j = ++i;
}
// prints 1 2 3 4

int i = 0;
int j = i;
while(j < 5) {
    printf("%d", i);
    j = i++;
}
// prints 0 1 2 3 4 5 
```

http://ideone.com/U3Wz2K

scripts/install/pcl_install.bash
Jebediah: It is from PCL's website: http://pointclouds.org/downloads/linux.html

Jebediah: I don't think we should use the one in ROS Kinetic since libwave is supposed to be ROS-free. We could build from source, but it makes travis slower.

.gitignore
Jebediah: These are for ignoring some garbage created by the IDE I am using (clion).

chutsu: quite a few lab members use clion so that addition to gitignore is fine. We currently don't have a universal `.gitignore`, I personally don't see the problem adding custom gitignore rules as they come (since they are pretty rare).

scripts/install/pcl_install.bash
stevenwaslander: OK.

.gitignore
stevenwaslander: OK.

wave_matching/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.3)
+PROJECT(wave_matching)
+
+IF (CMAKE_COMPILER_IS_GNUCXX)
+    ADD_DEFINITIONS(-std=gnu++11)
+ENDIF()
```
chutsu: this is defined in the master branch in root `CMakeFile.txt` so this is not needed.

chutsu: this is defined in the master branch in global `CMakeFile.txt` so this is not needed.

wave_matching/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
...
+        /usr/lib
+        /usr/local/lib)
+
+set(DEPS
+    ${PCL_COMMON_LIBRARIES}
+    ${PCL_REGISTRATION_LIBRARIES}
+    ${PCL_POINT_TYPES_LIBRARIES}
+    ${PCL_IO_LIBRARIES})
+
+add_library(wave_matching STATIC src/ICP.cpp)
```
chutsu: I believe in our style guide we say source files should all be lower case separated by an underscore.

wave_matching/tests/icp_tests.cpp
```diff
@@ -23,30 +25,34 @@ bool isApprox(const Eigen::Affine3d &first,
...
     virtual void SetUp() {
         this->ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
         this->target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
         pcl::io::loadPCDFile("../../wave_matching/tests/testscan.pcd",
                              *(this->ref));
     }
+
     void setParams(const float res, const Eigen::Affine3d perturb) {
-        this->matcher = new wave::ICP_Matcher(res);
+        this->matcher = new ICP_Matcher(res);
```
chutsu: No underscore

wave_matching/tests/icp_tests.cpp
```diff
@@ -23,30 +25,34 @@ bool isApprox(const Eigen::Affine3d &first,
 
 // Fixture to load same pointcloud all the time
 class ICPTest : public testing::Test {
-protected:
+ protected:
     ICPTest() {}
+
     virtual ~ICPTest() {
         if (matcher) {
```
chutsu: `this->`

wave_matching/tests/icp_tests.cpp
```diff
@@ -23,30 +25,34 @@ bool isApprox(const Eigen::Affine3d &first,
...
             delete matcher;
```
chutsu: `this->`

wave_matching/include/wave/matching/ICP.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <pcl/point_types.h>
+#include <pcl/point_cloud.h>
+#include <pcl/registration/icp.h>
+#include <pcl/filters/voxel_grid.h>
+#include "wave/matching/matcher.hpp"
+
+namespace wave {
+
+typedef pcl::PointCloud<pcl::PointXYZ>::Ptr PCLPointCloud;
+class ICP_Matcher : public wave::Matcher<PCLPointCloud> {
```
chutsu: under `wave` namespace already don't need wave here

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+#include <Eigen/Core>
+#include <Eigen/Geometry>
+
+namespace wave {
+
+// sensor data type
+template <typename T>
+class Matcher {
+ public:
```
chutsu: you didn't run clang-format on this file, else `public` would not be indented

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+// Zero displacement without downsampling
+TEST_F(ICPTest, fullResNullMatch) {
+    Eigen::Affine3d perturb;
+    Eigen::Affine3d result;
+    bool match_success = false;
+    // Setup
+    perturb = Eigen::Affine3d::Identity();
+    perturb.translation() << 0, 0, 0;
+    setParams(-1, perturb);
```
chutsu: `this->`

wave_matching/tests/icp_tests.cpp
```diff
@@ -0,0 +1,111 @@
...
+    EXPECT_TRUE(match_success);
+    EXPECT_TRUE(isApprox(result, perturb, 0.1));
+}
+
+}  // end of namespace wave
+
+int main(int argc, char *argv[]) {
+    testing::InitGoogleTest(&argc, argv);
+    return RUN_ALL_TESTS();
+}
```
chutsu: At current I have a general wave_utils_tests and wave_optimization_tests test runner for each module of libwave, maybe you should do the same? For example this is how I link up all my test cases into a single test runner in the utils module:

```
# UNIT TESTS
SET(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}/tests)
FILE(COPY tests/data DESTINATION ${PROJECT_BINARY_DIR}/tests)

ADD_EXECUTABLE(
    # TARGET
    wave_utils_tests
    # TESTS
    tests/utils/config_test.cpp
    tests/utils/data_test.cpp
    tests/utils/filesystem_test.cpp
    tests/utils/math_test.cpp
    tests/utils/time_test.cpp
    # TEST RUNNER
    tests/test_runner.cpp
)
TARGET_LINK_LIBRARIES(wave_utils_tests ${TEST_DEPS})
```
Open to suggestions if you have alternatives.

wave_matching/CMakeLists.txt
Jebediah: ok

wave_matching/include/wave/matching/matcher.hpp
Jebediah: yes I "fixed" that part manually. Google style guide says to have a 1-space indent on labels inside classes.

wave_matching/tests/icp_tests.cpp
Jebediah: I like this. It will scale well with more matchers.

wave_matching/include/wave/matching/matcher.hpp
chutsu: I think we might want to have an exception to that rule. I'm in favor of no indent, unless you have problems with it? 

@michaelhsmart? thoughts?

chutsu: ok mike says nay.

Can you change the `.clang-format` rule from `AccessModifierOffset: -4` to `AccessModifierOffset: -3`

Jebediah: no problem, I couldn't care less about how many spaces a label gets. Exception just needs to be documented.

chutsu: We may revisit this later :D

michaelhsmart: commenting here so it's tracked:

We've used 1 for autonomoose to keep cpplint from spewing piles except where we truly do deviate.

Depends solely on your desired level of cpplint adherence.

chutsu:
When merging please edit the commit bullet points to only KEY ONES and select "Squash and Merge"

PR #29, opened 2017-03-28
Feature 23 litdocs

stevenwaslander:
Revised the file structure and laid out the main topics to be covered in the reference docs.  Also organized the open source review a bit, and move the component reviews to the respective reference docs files. So the goal structure is as follows, as defined in the sidebar.
1) API - code focused docs, should have the same main categories as fundamentals, many currently missing, but that's ok.
2) Fundamentals - breaks into Math, Kinematics, Vision, Lidar, Inertial, Estimation, Optimization
3) Required reading - breaks into Great SLAM papers, Open Source Code Reviews and Reading course content.  

Notes:
1) Each fundamental topic should start with any solid summaries available elsewhere, or end with it.
2) The surveys and complete SLAM papers that we are comparing to did not fit in fundamental sub categories, so they get their own list.
3) The reading course content should actually be in the fundamentals topics, but this will be for slides and videos that go over each area.  It will likely be duplicate links to do it this way, and could ultimately be moved and linked into each of the fundamental topics.  Ultimately, I see the reading course disappearing except for in Basecamp, but I've kept it for now.


stevenwaslander:
What? I don't understand how building could fail. Merged everything in from Master, and only changed docs.

Issue #26, opened 2017-03-23
Documentation for Utils


arunxdas:
Utils needs documentation.  Top priority are the wave:: typedefs (Mat3, Vec3, etc), as these are used throughout other modules.

PR #30, opened 2017-03-29
Docs 26 utils

chutsu:
This is in relation to issue #26:

In summary this PR has the following changes:

- Add docs for `wave_utils`
- Refactor `ConfigParam` to have a generic `void *` instead of a long list of specific pointers for each type
- Cleaned up error codes in `ConfigParser`


stevenwaslander:
Let's pick 2 or at most 3 reviewers per PR, so we don't get overwhelmed, and every request is honoured.

wave_utils/src/config.cpp
```diff
@@ -144,31 +247,33 @@ int ConfigParser::loadArray(ConfigParam param) {
     switch (param.type) {
         case BOOL_ARRAY:
             for (auto n : node) {
-                param.b_array->push_back(n.as<bool>());
+                ((std::vector<bool> *) param.data)->push_back(n.as<bool>());
```
leokoppel: Use `static_cast`

wave_utils/src/config.cpp
```diff
@@ -117,12 +215,12 @@ int ConfigParser::loadPrimitive(ConfigParam param) {
     this->getYamlNode(param.key, node);
     // clang-format off
     switch (param.type) {
-        case BOOL: *param.b = node.as<bool>(); break;
-        case INT: *param.i = node.as<int>(); break;
-        case FLOAT: *param.f = node.as<float>(); break;
-        case DOUBLE: *param.d = node.as<double>(); break;
-        case STRING: *param.s = node.as<std::string>(); break;
-        default: return -3;
+        case BOOL: *(bool *) param.data = node.as<bool>(); break;
```
leokoppel: Use `static_cast`

wave_utils/src/config.cpp
```diff
@@ -4,15 +4,97 @@
 namespace wave {
 
 ConfigParser::ConfigParser(void) {
-    this->configured = false;
-    this->loaded = false;
+    this->config_loaded = false;
 }
 
+// clang-format off
```
chutsu: Nope, I'm good with 2 space indent :D

PR #31, opened 2017-03-31
Feature 8 pcl gicp

Jebediah:
This is to wrap PCL's G-ICP matcher. See #7 and #8 .
Pretty much equivalent to #28 but for G-ICP


wave_matching/config/g_icp.yaml
```diff
@@ -0,0 +1,5 @@
+g_icp:
+  correspondenceRandomness: 10
```
chutsu: lets not mix up the cases, lets make the config reflect that class member variables 1-to-1 mapping to avoid confusion. So in this case this line should be `corr_rand: 10`.

wave_matching/src/g_icp.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+    double r_eps = 1e-8, fit_eps = 1e-2;
+    int corr_rand = 10, max_iter = 100;
+
+    parser.addParam("g_icp.correspondenceRandomness", &corr_rand);
+    parser.addParam("g_icp.maxIterations", &max_iter);
+    parser.addParam("g_icp.rotationEpsilon", &r_eps);
+    parser.addParam("g_icp.euclideanFitnessEpsilon", &fit_eps);
+
+    parser.load("config/g_icp.yaml");
```
Jebediah: If there is an error, what should I do about it?

chutsu: throw an exception

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -1,10 +1,17 @@
...
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 
 namespace wave {
+// define exception
+class ConfigException : public __exception {
+    virtual const char* what() const throw() {
+        return "Failed to Load Matcher Config";
+    }
+};
```
arunxdas: Are we looking to define custom exceptions? or try and stick with standard ones.  What's the downside to either argument?

Jebediah: I looked at http://www.cplusplus.com/doc/tutorial/exceptions/ for the standard exceptions. The closest one on there is "runtime_error" which doesn't seem very helpful.

chutsu: I personally think in this instance you should define a specific "MatcherException".

Issue #9, opened 2017-03-13
PCL NDT (voxels)


chutsu:

Jebediah:
Subissue of #7

arunxdas:
Is this the wrapping of PCL's NDT scan matcher?

Jebediah:
That's the plan.

PR #35, opened 2017-04-04
Feature 9 pcl ndt

Jebediah:
This PR is to wrap PCL's 3D-NDT matcher. See #9  and #7. Equivalent to #28 / #31 .

Note that I changed the build configuration to release to get the tests to run faster


wave_matching/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 2.8.3)
 PROJECT(wave_matching)
 
+set( CMAKE_BUILD_TYPE release )
+
 find_package(PCL REQUIRED common registration point_types io)
 find_package(OpenCV REQUIRED)
```
chutsu: Make your CMAKE commands CAPITALS, e.g. `SET(....), FIND_PACKAGE(...)`, whether its lower or upper cmake doesn't care but for consistency lets make it caps, it also stands out so we know what is CMAKE what is a file/target

wave_matching/tests/ndt_tests.cpp
```diff
@@ -53,7 +53,7 @@ TEST_F(NDTTest, fullResNullMatch) {
     match_success = matcher->match();
     double diff = (matcher->getResult().matrix() - perturb.matrix()).norm();
     EXPECT_TRUE(match_success);
-    EXPECT_LT(diff, 0.1);
+    EXPECT_LT(diff, 0.12);
```
chutsu: If I'm not mistaken actually, are GTests Assert arguments `ASSERT(expected, actual)`?

Jebediah: For these comparisons it's as if the operator is inserted between the arguments. 
EXPECT_LT( val1, val2 ) checks if val1 < val2. ([gtest](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md))

chutsu: ah right! me being stupid..

wave_matching/include/wave/matching/ndt.hpp
```diff
@@ -0,0 +1,29 @@
...
+ public:
+    explicit NDTMatcher(float resolution, const std::string &config_path);
+    void setRef(const PCLPointCloud &ref);
+    void setTarget(const PCLPointCloud &target);
+    bool match();
+
+ private:
+    pcl::NormalDistributionsTransform<pcl::PointXYZ, pcl::PointXYZ> ndt;
+    PCLPointCloud ref, target, final;
+    const float min_res = 0.05f;
```
Jebediah: Probably not

wave_matching/tests/ndt_tests.cpp
```diff
@@ -0,0 +1,91 @@
...
+    perturb.translation() << 0.2, 0, 0;
+    this->setParams(2.5f, perturb);
+    // test and assert
+    match_success = matcher->match();
+    double diff = (matcher->getResult().matrix() - perturb.matrix()).norm();
+    EXPECT_TRUE(match_success);
+    EXPECT_LT(diff, 0.1);
+}
+
+}  // end of namespace wave
```
Jebediah: ? Where's the symbol?

wave_matching/src/icp.cpp
```diff
@@ -13,17 +13,20 @@ ICPMatcher::ICPMatcher(float res) {
     } else {
         this->resolution = -1;
     }
-    wave::ConfigParser parser;
+    ConfigParser parser;
 
     double max_corr = 3, t_eps = 1e-8, fit_eps = 1e-2;
```
Jebediah: Yeah that's left over from an earlier stage when these were all hard-coded. I will remove the initialisations.

wave_matching/src/ndt.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace wave {
+
+NDTMatcher::NDTMatcher(float res, const std::string &config_path) {
+    this->ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+    this->target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+    this->final = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+
+    ConfigParser parser;
+
+    float step_size = 0.1, t_eps = 0.01, config_res = 3;
```
Jebediah: k

wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -1,10 +1,17 @@
 #ifndef WAVE_MATCHING_MATCHER_HPP
 #define WAVE_MATCHING_MATCHER_HPP
 
+#include <exception>
 #include <Eigen/Core>
 #include <Eigen/Geometry>
 
 namespace wave {
+// define exception
```
chutsu: Superfluous comment

wave_matching/include/wave/matching/ndt.hpp
```diff
@@ -0,0 +1,29 @@
+// Class to wrap PCL ICP matcher
```
chutsu: Superfluous comment, the first line in a header should be a define anyways

wave_matching/include/wave/matching/ndt.hpp
chutsu: In this case the "f" makes sense because its explicity saying this float is of type `float`, without it by default its a `double` and if you do `float min_res = 0.05` you are implicitly casting the double to float, small detail but this case it makes perfect sense to have the "f" at the end.

wave_matching/src/ndt.cpp
```diff
@@ -0,0 +1,63 @@
+#include "wave/utils/config.hpp"
+#include "wave/utils/logging.hpp"
+#include "wave/matching/ndt.hpp"
+
+namespace wave {
+
+NDTMatcher::NDTMatcher(float res, const std::string &config_path) {
+    this->ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+    this->target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+    this->final = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
```
chutsu: Are you using the clang-format? I don't remember the settings allowing spaces at the last angle bracket ">"

wave_matching/tests/icp_tests.cpp
```diff
@@ -35,13 +21,13 @@ class ICPTest : public testing::Test {
     }
 
     virtual void SetUp() {
-        this->ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
-        this->target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        this->ref = boost::make_shared < pcl::PointCloud < pcl::PointXYZ > > ();
+        this->target = boost::make_shared < pcl::PointCloud < pcl::PointXYZ > > ();
```
chutsu: This format seems inconsistent?

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+        pcl::ApproximateVoxelGrid<pcl::PointXYZ> filter;
+        PCLPointCloud ref, target, final;
+        const float min_res = 0.05f;
+    };
+
+
+### Parameters
+
+There are a few parameters that may be changed specific to this algorithm. They
+can be set in the yaml config file.
```
arunxdas: specify where the yaml file lives, unless it's some kind of standard location

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+
+* step_size: Maximum Newton step size used in line search
+* max_iter: Limits number of iterations
+* t_eps: Criteria to stop iterating. If the difference between consecutive transformations is less than this, stop.
+* config_res: If the contructor is given an invalid resolution (too fine or negative), use this resolution
```
arunxdas: is default_res a more intuitive name?

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Constructor
+
+    explicit NDTMatcher(float resolution, const std::string &config_path);
```
arunxdas: float &resolution for input

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+        explicit NDTMatcher(float resolution, const std::string &config_path);
+        void setRef(const PCLPointCloud &ref);
+        void setTarget(const PCLPointCloud &target);
+        bool match();
+
+     private:
+        pcl::NormalDistributionsTransform<pcl::PointXYZ, pcl::PointXYZ> ndt;
+        pcl::ApproximateVoxelGrid<pcl::PointXYZ> filter;
+        PCLPointCloud ref, target, final;
+        const float min_res = 0.05f;
```
arunxdas: do you need the f at the end if it's a float type?

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+
+    typedef pcl::PointCloud<pcl::PointXYZ>::Ptr PCLPointCloud;
+
+    }  // end of wave namespace
+
+This type definition is used as a shorthand for the pointcloud object type used
+by the scan matching implementations in PCL. A number of those implementations are
+wrapped, so they use the same datatype for the pointcloud.
+
+## ICP Matcher
```
arunxdas: NDT Matcher?

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Constructor
+
+    explicit NDTMatcher(float resolution, const std::string &config_path);
+
+This constructor takes an argument in order to specify resolution. The resolution
+given takes precedence over the one in the config file. If both resolutions are
+finer than the min_res class member, the resolution is set to min_res.
+
+config_path is the path to the yaml config file
```
arunxdas: period at end

docs/api/matching/ndt.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Private Members
+
+    pcl::NormalDistributionsTransform<pcl::PointXYZ, pcl::PointXYZ> ndt;
+
+This is an instance of the NDT class from PCL
+
+---
+
+    PCLPointCloud ref, target, final;
```
arunxdas: are these pointers? shouldn't it be pcl::PointCloud::Ptr or something?

wave_matching/include/wave/matching/ndt.hpp
```diff
@@ -0,0 +1,29 @@
...
+#include <pcl/point_cloud.h>
+#include <pcl/registration/ndt.h>
+#include "wave/matching/matcher.hpp"
+#include "wave/matching/pcl_common.hpp"
+
+namespace wave {
+
+class NDTMatcher : public Matcher<PCLPointCloud> {
+ public:
+    explicit NDTMatcher(float resolution, const std::string &config_path);
```
arunxdas: const float &resolution for input

wave_matching/include/wave/matching/ndt.hpp
arunxdas: does the pcl resolution only accept float as a type? it won't take a double?

wave_matching/src/icp.cpp
```diff
@@ -13,17 +13,20 @@ ICPMatcher::ICPMatcher(float res) {
     } else {
         this->resolution = -1;
     }
-    wave::ConfigParser parser;
+    ConfigParser parser;
```
arunxdas: can we also make another constructor that takes in a subset of the important params, instead of the full config file.  In some cases we may have multiple instantiations of this object with different parameters, and I can see it getting annoying to have different config files for each object, or use the one default config file then manually call the "set" methods.

wave_matching/src/ndt.cpp
```diff
@@ -0,0 +1,63 @@
+#include "wave/utils/config.hpp"
+#include "wave/utils/logging.hpp"
+#include "wave/matching/ndt.hpp"
+
+namespace wave {
+
+NDTMatcher::NDTMatcher(float res, const std::string &config_path) {
```
arunxdas: same as the ICP comment, make constructor that takes in a subset of the most important params, default the rest.  That way if we want multiple matchers with different params (think multi-scale optimization), we don't need a bunch of different config files, or have to call set methods all the time

wave_matching/src/ndt.cpp
```diff
@@ -0,0 +1,63 @@
...
+    parser.addParam("ndt.config_res", &config_res);
+
+    if (parser.load(config_path) != 0) {
+        ConfigException config_exception;
+        throw config_exception;
+    }
+
+    if (res < this->min_res) {
+        if(config_res < this->min_res) {
+            LOG_ERROR("Invalid resolution given, using minimum");
```
arunxdas: What is the LOG_ERROR macro?

wave_matching/tests/icp_tests.cpp
```diff
@@ -65,9 +51,9 @@ TEST_F(ICPTest, fullResNullMatch) {
     this->setParams(-1, perturb);
     // test and assert
     match_success = matcher->match();
-    result = matcher->getResult();
+    double diff = (matcher->getResult().matrix() - perturb.matrix()).norm();
     EXPECT_TRUE(match_success);
-    EXPECT_TRUE(isApprox(result, perturb, 0.1));
+    EXPECT_LT(diff, 0.1);
```
arunxdas: make the 0.1 a member function of your test fixture, some kind of threshold or something

arunxdas: In the future, once the Transformation class is done, you should be comparing two matrices using the supplied "isNear" function.  Can you make a ticket for this so we don't forget?

wave_matching/tests/icp_tests.cpp
```diff
@@ -81,9 +67,9 @@ TEST_F(ICPTest, nullDisplacement) {
     this->setParams(0.05f, perturb);
     // test and assert
     match_success = matcher->match();
-    result = matcher->getResult();
+    double diff = (matcher->getResult().matrix() - perturb.matrix()).norm();
     EXPECT_TRUE(match_success);
-    EXPECT_TRUE(isApprox(result, perturb, 0.1));
+    EXPECT_LT(diff, 0.1);
```
arunxdas: same comment as above, change 0.1 to a threshold defined in your test fixture.  Change throughout.

wave_matching/tests/ndt_tests.cpp
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    virtual void SetUp() {
+        this->ref = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        this->target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
+        pcl::io::loadPCDFile("data/testscan.pcd", *(this->ref));
+    }
+
+    void setParams(const float res, const Eigen::Affine3d perturb) {
```
arunxdas: this does a lot more than just setParams.  Perhaps "initializePointcloud" or something is a better name?

wave_matching/tests/ndt_tests.cpp
arunxdas: yes, should add newline

docs/api/matching/ndt.md
chutsu: yes, its more explicit, without it C++ treats it as a double and will attempt to cast double to float.

docs/api/matching/ndt.md
chutsu: you sure? should only do a pass by reference only if you want the variable `resolution` outside the scope of `NDTMatcher` to change. Plus general rule of thumb always pass objects by reference `&`, but never on built-in types (aka primitives; `int`, `float`, `double`, `size_t`, etc.)

wave_matching/include/wave/matching/ndt.hpp
chutsu: Not required, passing by value for built-in types are generally preferred, further const does not make sense because you are passing by value, `const` keyword does nothing.

wave_matching/src/ndt.cpp
chutsu: The general error logging macro, it is defined [here](https://github.com/wavelab/libwave/blob/master/wave_utils/include/wave/utils/logging.hpp)

chutsu: Obviously we can change this to something else, but for now it is sufficient

wave_matching/src/ndt.cpp
Jebediah: Ran clang, didn't change it.

docs/api/matching/ndt.md
michaelhsmart: Arun probably meant "`const float& resolution` for input"

but ya, pass by value for built-in types is fine

wave_matching/include/wave/matching/ndt.hpp
arunxdas: right,but I was thinking more towards readability, where if you pass by ref and const it, then it's fairly clear that it's an input to the function and it won't be modified.  Why is pass by value preferred in this case?

chutsu: The general consensus is that while pass by reference will avoid copying, small built-in types such as `float` will have negligible impact on performance in contrast to copying potentially large objects.

chutsu: In regards to whether you should do `const float` or simply just `float` I leave that up to you, I personally don't think the `const` here adds anything because I already know I'm passing by value therefore there is no chance `resolution` will change. Up to Ben :P

docs/api/matching/ndt.md
Jebediah: There is a typedef in pcl_common.hpp. 
`typedef pcl::PointCloud<pcl::PointXYZ>::Ptr PCLPointCloud;`

docs/api/matching/ndt.md
arunxdas: yes, meant const float &resolution, fixed above.  This is good discussion, I didn't know until now that there are some real optimisation and speed benefits to passing primitives by value (according to stack exchange)

http://gph.is/15qhNmo

wave_matching/include/wave/matching/ndt.hpp
arunxdas: agreed, and also apparently there are some performance costs to passing primitives by reference.  Top answer here: http://stackoverflow.com/questions/14013139/is-it-counter-productive-to-pass-primitive-types-by-reference

wave_matching/src/icp.cpp
Jebediah: Voxel resolution is already set via the constructor, the rest are parameters like maximum iterations, stopping criteria, maximum correspondences, etc. Are these really the type of parameters you would want to change between scan matches?
Multiple config files might be better for something like multi-scale optimisation because they expose all the configuration parameters without having to look at code.

wave_matching/src/ndt.cpp
Jebediah: something from wave utils [logging](https://github.com/wavelab/libwave/blob/master/wave_utils/include/wave/utils/logging.hpp). Prints error to stderr.

wave_matching/tests/ndt_tests.cpp
Jebediah: I'll change to initMatcher

wave_matching/include/wave/matching/ndt.hpp
Jebediah: It will take a double in the sense that there will be an implicit conversion to a float.

wave_matching/src/icp.cpp
arunxdas: that's fair, leave it for now and if we need to add more constructors later, we can do that.

wave_matching/src/ndt.cpp
arunxdas: see above comment

Issue #12, opened 2017-03-15
Kinematics-Rotation


arunxdas:
implement the rotations component of the kinematics library

PR #38, opened 2017-04-05
Add quadrotor model from prototype

chutsu:
Copying over the Quadrotor motion model I used in my Gazebo Sim. At current it has hard coded PID tunes within the Quadrotor class, but future work can improve on that small detail. For now I just want to add this in.


stevenwaslander:
First step in fitting this into the libwave structure.  Separated kinematics from controls libraries (at least partially).  The kinematics still has the actual quadrotor controllers in it, which I want to separate from the quadrotor model that should be in kinematics, but as a first step I rejiggered CMakeLists and added libraries and fixed a few bugs along the way (e.g. VelocityController test was still in there, although VelocityController was not.)
Had to add in euler2rot from awesomo's math.cpp, but I assume there was a reason that was not included in the transfer.  I'd appreciate hearing how I should do this properly, but wanted to push something that worked.
Please have a look, and comment away.  Obviously not a finished PR, but as advisor, I get to ask for help this way! #Abuseofpower.

chutsu:
LGTM, I only created the new `wave_quadrotor` module because I was hoping it wouldn't conflict with Arun's kinematics module. I had planned to put quadrotor code into kinematics later once feature #12 is merged in, but either way works.

`euler2rot()` was removed because eventually the kinematics module will have that function, just that my old code still used it. I suppose when the time comes we will replace this with the Arun's version at a later date.

stevenwaslander:
OK thanks, just leave this one for now, we should get the kinematics branch review done first, then I can cycle back and try to bring this one in line.

wave_kinematics/include/wave/kinematics/quadrotor.hpp
```diff
@@ -0,0 +1,86 @@
...
+ public:
+    double dt;
+    Vec4 outputs;
+
+    PID roll_controller;
+    PID pitch_controller;
+    PID yaw_controller;
+
+    AttitudeController(void);
+    Vec4 update(Vec4 setpoints, Vec4 actual, double dt);
```
leokoppel: vector parameters should be `const &` (for all methods)

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+    this->dt = 0.0;
+
+    return outputs;
+}
+
+
+// QUADROTOR MODEL
+QuadrotorModel::QuadrotorModel(void) {
+    this->states = VecX(12);
+    this->states(0) = 0.0;   // roll
```
leokoppel: Why is `states` a vector? It's always used one element at a time, and you have to remember which position is which.
Split into individual members, or 3d vectors for position, velocity, etc.

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+    this->states(4) = 0.0;   // angular velocity y
+    this->states(5) = 0.0;   // angular velocity z
+    this->states(6) = 0.0;   // x
+    this->states(7) = 0.0;   // y
+    this->states(8) = 0.0;   // z
+    this->states(9) = 0.0;   // linear velocity x
+    this->states(10) = 0.0;  // linear velocity y
+    this->states(11) = 0.0;  // linear velocity z
+
+    this->Ix = 0.0963;  // inertial x
```
leokoppel: Put these comments into the header (or, use non-abbreviated variable names).
Put the initialization into the header as well.
Some of these should be `const`.

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+    this->g = 10.0;  // gravitational constant
+
+    this->attitude_setpoints = Vec4();
+    this->position_setpoints = VecX(3);
+
+    this->attitude_controller = AttitudeController();
+    this->position_controller = PositionController();
+}
+
+QuadrotorModel::QuadrotorModel(VecX pose) {
```
leokoppel: No need to duplicate the whole constructor.

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+    this->position_setpoints = VecX(3);
+    this->position_setpoints << pose(0), pose(1), pose(2);
+
+    this->attitude_controller = AttitudeController();
+    this->position_controller = PositionController();
+}
+
+int QuadrotorModel::update(VecX motor_inputs, double dt) {
+    double ph, th, ps;
```
leokoppel: ?

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+  this->states(0) = ph + (p + q * sin(ph) * tan(th) + r * cos(ph) * tan(th)) * dt;
+  this->states(1) = th + (q * cos(ph) - r * sin(ph)) * dt;
+  this->states(2) = ps + ((1 / cos(th)) * (q * sin(ph) + r * cos(ph))) * dt;
+  this->states(3) = p + (-((Iz - Iy) / Ix) * q * r - (kr * p / Ix) + (1 / Ix) * taup) * dt;
+  this->states(4) = q + (-((Ix - Iz) / Iy) * p * r - (kr * q / Iy) + (1 / Iy) * tauq) * dt;
+  this->states(5) = r + (-((Iy - Ix) / Iz) * p * q - (kr * r / Iz) + (1 / Iz) * taur) * dt;
+  this->states(6) = x + vx * dt;
+  this->states(7) = y + vy * dt;
+  this->states(8) = z + vz * dt;
+  this->states(9) = vx + ((-kt * vx / m) + (1 / m) * (cos(ph) * sin(th) * cos(ps) + sin(ph) * sin(ps)) * tauf) * dt;
```
leokoppel: Mention what equation you're using?
Seems like it could be simplified by having 3d vectors for velocity and position, etc. and operating on those with a rotation matrix (assuming that's what is happening here)

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+    // constrain yaw to be [-180, 180]
+    this->states(2) = rad2deg(this->states(2));
+    this->states(2) = wrapTo180(this->states(2));
+    this->states(2) = deg2rad(this->states(2));
+
+    return 0;
+}
+
+VecX QuadrotorModel::attitudeControllerControl(double dt) {
```
leokoppel: Why `VecX` when it always returns `Vec4`? Same for other methods.

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+AttitudeController::AttitudeController(void) {
+    this->dt = 0.0;
+    this->outputs << 0.0, 0.0, 0.0, 0.0;
+
+    this->roll_controller = PID(200.0, 0.5, 10.0);
+    this->pitch_controller = PID(200.0, 0.5, 10.0);
+    this->yaw_controller = PID(200.0, 0.5, 10.0);
+}
+
+Vec4 AttitudeController::update(Vec4 setpoints, Vec4 actual, double dt) {
```
leokoppel: What do the 4 components of the vectors mean? Is it documented somewhere?

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+    this->l = 0.9;  // arm length
+    this->d = 1.0;  // drag
+
+    this->m = 1.0;   // mass of quad
+    this->g = 10.0;  // gravitational constant
+
+    this->attitude_setpoints = Vec4();
+    this->position_setpoints = VecX(3);
+
+    this->attitude_controller = AttitudeController();
```
leokoppel: It was already initialized before this line.

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+
+    x = this->states(6);
+    y = this->states(7);
+    z = this->states(8);
+
+    vx = this->states(9);
+    vy = this->states(10);
+    vz = this->states(11);
+
+    Ix = this->Ix;
```
leokoppel: It looks like these assignments are just to get around the extra length of `this->` huh?
Maybe change the rules to something convenient, instead of going to great lengths to circumvent them ;-) - -same for clang-format directives on long lines.

Or, to be clear about what you're doing
```cpp
const auto &Ix = this->Ix`;
```

This also works to keep calculations readable while having meaningful,  non-abbreviated member names.

wave_kinematics/src/quadrotor.cpp
```diff
@@ -0,0 +1,435 @@
...
+    this->position_controller = PositionController();
+}
+
+int QuadrotorModel::update(VecX motor_inputs, double dt) {
+    double ph, th, ps;
+    double p, q, r;
+    double x, y, z;
+    double vx, vy, vz;
+
+    double Ix, Iy, Iz;
```
leokoppel: http://stackoverflow.com/a/3773458/431033
and use auto

wave_kinematics/tests/quadrotor_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+        fprintf(output_file, "%f, ", quad.states(6));
+        fprintf(output_file, "%f, ", quad.states(7));
+        fprintf(output_file, "%f\n", quad.states(8));
+        t += dt;
+    }
+
+    fclose(output_file);
+}
+
+/*TEST(VelocityController, update) {
```
leokoppel: Remove dead code

wave_kinematics/tests/quadrotor_test.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include "wave/kinematics/quadrotor.hpp"
+#include "wave/controls/pid.hpp"
+
+namespace wave {
+
+#define ATTITUDE_CONTROLLER_OUTPUT "/tmp/quadrotor_attitude_controller.dat"
+#define POSITION_CONTROLLER_OUTPUT "/tmp/quadrotor_position_controller.dat"
+#define VELOCITY_CONTROLLER_OUTPUT "/tmp/quadrotor_velocity_controller.dat"
+
+TEST(AttitudeController, update) {
```
leokoppel: Test doesn't test anything

wave_kinematics/tests/quadrotor_test.cpp
chutsu: Done

wave_controls/include/wave/controls/pid.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef WAVE_PID_HPP
+#define WAVE_PID_HPP
+
+#include <float.h>
```
chutsu: Removed

wave_controls/include/wave/controls/pid.hpp
```diff
@@ -0,0 +1,28 @@
...
+#include <math.h>
```
chutsu: Removed

wave_controls/src/pid.cpp
```diff
@@ -0,0 +1,51 @@
+#include "wave/controls/pid.hpp"
+
+namespace wave {
+
+PID::PID(void) {
+    this->error_prev = 0.0;
```
chutsu: Matter of taste, what difference does it really make?

wave_controls/src/pid.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+PID::PID(double k_p, double k_i, double k_d) {
+    this->error_prev = 0.0;
+    this->error_sum = 0.0;
+
+    this->error_p = 0.0;
+    this->error_i = 0.0;
+    this->error_d = 0.0;
+
+    this->k_p = k_p;
```
chutsu: Same comment as above, matter of taste, semantically changes nothing. Additionally a performance argument does not apply here since this class will only be declared a hand full of times .

wave_controls/src/pid.cpp
```diff
@@ -0,0 +1,51 @@
...
+    this->k_i = k_i;
+    this->k_d = k_d;
+}
+
+double PID::update(double setpoint, double actual, double dt) {
+    double error;
```
chutsu: This will require more discussion, as I am not sure the AAA style is a good rule to follow. It is indeed very useful for-range loops, making containers declarations shorter and templating. But the overuse of `auto` will at some point hinder readability when types do matter, given that our team is not at as high level of proficiency at C++ as you are I regard the overuse of AAA style a disadvantage rather than an advantage. Again open to discussion if you have strong opinions about this.

wave_controls/tests/control_test.cpp
```diff
@@ -0,0 +1,63 @@
+#include "wave/wave_test.hpp"
+#include "wave/controls/pid.hpp"
+
+namespace wave {
+
+TEST(PID, constructor) {
+    PID controller;
+
+    ASSERT_FLOAT_EQ(0.0, controller.error_prev);
```
chutsu: Done

wave_controls/tests/control_test.cpp
```diff
@@ -0,0 +1,63 @@
...
+    // setup
+    controller = PID(1.0, 1.0, 1.0);
+
+    // test and assert
+    output = controller.update(10.0, 0.0, 0.1);
+
+    ASSERT_FLOAT_EQ(1.0, controller.error_sum);
+    ASSERT_FLOAT_EQ(10.0, controller.error_p);
+    ASSERT_FLOAT_EQ(1.0, controller.error_i);
+    ASSERT_FLOAT_EQ(100.0, controller.error_d);
```
chutsu: Not certain, but I think it depends on where the PID is used, hence its implementation specific. I wanted this PID class to be as barebones as possible, one can introduce safeguards towards the `I`-term for example independently, etc.

wave_controls/include/wave/control.hpp
```diff
@@ -0,0 +1,28 @@
+#ifndef WAVE_CONTROL_HPP
```
chutsu: `control.hpp` removed

wave_kinematics/src/quadrotor.cpp
chutsu: Sorry this is my bad this code was rushed in as a side project for steve, and I didn't give it thought to  explan the code well enought. Still, no excuses, will add documentation soon.

wave_controls/src/pid.cpp
leokoppel: Fewer lines of code (especially when you have multiple, slightly different, constructors as here), less likely to forget to initialize something.

wave_controls/src/pid.cpp
leokoppel: You're right there is no difference. It would only be for consistency with cases where initializer lists are required.

wave_controls/src/pid.cpp
leokoppel: I do. If anything AAA makes it easier to code without accidentally uninitialized variables and subtle conversion bugs. Meyers addresses this in _Effective Modern C++_ ("Item 5: Prefer auto to explicit type declarations") , with some examples.

> The fact of the matter is that writing types explicitly often does little more than intro‐
duce opportunities for subtle errors, either in correctness or efficiency or both.

Herb Sutter also has a good response to the "readability" argument here:
https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

There are cases where you are right to be concerned and naive `auto` use is not good, in particular with Eigen expressions. (see _Effective Modern C++_ Item 6.) AAA can still be helpful here, but it does require further discussion.

wave_kinematics/src/quadrotor.cpp
chutsu: Refactored to Vec4

wave_kinematics/include/wave/kinematics/quadrotor.hpp
chutsu: Done

wave_kinematics/tests/quadrotor_test.cpp
chutsu: Done

wave_controls/src/pid.cpp
chutsu: Refactored so that all classes in this PR uses intiailizer lists.

wave_controls/src/pid.cpp
chutsu: Done

Issue #41, opened 2017-04-05
Add measurement containers


leokoppel:
Placeholder issue for measurement container feature.

Discussion at https://3.basecamp.com/3692715/buckets/3007033/documents/441104639

chutsu:
I think the main features have been implemented, close?

leokoppel:
See the basecamp link for list of features. #82 will close it when it is merged.

PR #42, opened 2017-04-07
Add measurement containers

leokoppel:
Partial implementation of the  [measurement containers interface](https://3.basecamp.com/3692715/buckets/3007033/documents/441104639) (#41). Includes insertion, deletion, and retrieval with interpolation of one element at a time, and iteration over the
whole container.

I changed some of the method names and return types to be consistent with standard containers.

Questions
- Where to put this class? (in utils directory for now)
- For this class template, should methods go inside the class body or outside? Outside (as they are now) may make the declarations cleaner but adds a lot of bloat, especially for single-line methods.


leokoppel:
It looks like Travis has Boost 1.54, but a number of C++11-compliant methods were only added in 1.55. I will add support for 1.54.

stevenwaslander:
It seems there should be a wave-core subfolder for measurement and state containers, or these could go in wave-estimation.  Utils should be more for utilities, and less for core elements like this. (Opinion only, not law).

wave_utils/include/wave/utils/measurement.hpp
```diff
@@ -0,0 +1,312 @@
...
+            ordered_unique<tag<composite_index>, sensor_and_time_key>> {
+    };
+    // Note we use separate struct definitions above, instead of typedefs, to
+    // reduce the length of the name printed by the compiler.
+
+    // Finally, define the multi_index_container type.
+    using type = boost::multi_index_container<T, indices, std::allocator<T>>;
+
+    // Get the type of the composite index, using its tag
+    using composite_type = typename type::template index<composite_index>::type;
```
StanJBrown: What on earth is this doing? I was just browsing out of interest, but this code is way beyond my understanding. I feel like this will be an issue going forward with code maintenance.

wave_utils/include/wave/utils/measurement.hpp
```diff
@@ -0,0 +1,312 @@
...
+            value{v} {}
+};
+
+/** Perform linear interpolation between two measurements
+ *
+ * This template can be specialized for measurement types requiring a
+ * different calculation
+ */
+template<typename T>
+decltype(T::value) interpolate(const T &m1, const T &m2, const TimeType &t) {
```
StanJBrown: decltype? What dees that do? Also for interpolate, should this indicate that it is purely linear interpolation?

StanJBrown:
I was just taking a break my thesis writing for fun, and at a high level, I find this code very hard to understand. I feel like the only people who could completely  explain it to me are Leo or Chris. I am personally not familiar with boost. Would it be worth while having a set of docs at least explaining how some of these classes with things like boost multiindex, and why we are using them here. What is the advantage of them?

leokoppel:
Hi Stan, those are good points. I should improve the comments. 

The MeasurementContainer class wraps the [Boost MultiIndex library](http://www.boost.org/doc/libs/1_63_0/libs/multi_index/doc/index.html). It lets us store entries and access them using multiple kinds of keys: in this case we can have them sorted by time, or sorted by sensor ID, or both.  The MultiIndex container is configured via template parameters, which is the messy type definitions you noticed.

I would like to point out that this code—that is, the code I wrote which we have to maintain, not the underlying library—is not complex. IMO it is the simplest and most boring solution for the given task. Note that this entire PR contains about 5 lines of runtime logic. I would be a lot more concerned if I opened a PR and saw a couple of nested loops and some pointer math.

That said, I agree with your concerns about readability, and I'll try to improve the layout and comments / documentation so it is understandable and maintainable.

michaelhsmart:
I'm just going to comment here on this discussion.

Independent of the underlying Boost library usage, templating usage here needs some attention. Stan mentioned decltype usage, but I'm going to extrapolate on his comment to aim for what I think he's really getting at - template metaprogramming (TMP) can be hard to read. 

Given what you guys want the MeasurementContainer class to do, this use of templating is entirely appropriate IMO. I don't generally like TMP as I find it can quickly become a bit arcane and difficult to follow, but for a Container class like what you guys want here, especially with these type-abstract functions you want supported, you'd be crazy to not use it.

That being said, TMP is easily the weakest skill-set we have in the lab. I'm not sure if documenting every line of TMP code in this class that might be slightly arcane to some is appropriate as we'd quickly end up with just as much commenting as we would code.

In addition to whatever API documentation you have in mind, it might be worth considering having separate implementation documentation that can give this kind of coverage of the basic TMP concepts/tricks that you are using here to implement the class without bogging down any of the core API or commenting to satisfy readers who've never seen TMP.

My general recommendation is that if someone who has our baseline familiarity would benefit from an illuminating comment, then use an illuminating comment. If this leads you to needing to comment a concept, you probably want to pull that out into a doc somewhere instead of copying a programming how-to-XYZ blog into a comment block. Maybe just have comments for tricks that would help someone with a very basic TMP level, and docs for the rest.

It's really up to you guys on the slam project, but I'd highly recommend considering it.

TLDR: TMP magic is probably a very good idea for this particular class. You probably want to have somewhat separated documentation on this since our baseline TMP level is low.

leokoppel:
@michaelhsmart good points. Note there are two uses of templates here

1. The template parameters used to configure the internal boost multi_index_container
2. The template parameter used by `MeasurementContainer`

The first one looks complicated but is just a use of a well-documented Boost interface. Boost may use TMP but on our side it is just defining a few types. It is an internal implementation detail, though - does it still belong in our documentation?

The second, I do think it needs better documentation on my side, and some examples, because it is an external interface. (Actually in the next PR I plan to put in measurements stored as Eigen types).

I would not call this "template metaprogramming", though. It is just using templates, or "generic programming".

michaelhsmart:
You can exchange "generic programming" for TMP in my comment. I'm more proficient in the code itself than I am with the terminology surrounding it.

My main point was that what you think of as clear template usage here is probably not going to be followed by a lot of the group (present and future). This is why I was thinking it might be appropriate to separate the documentation here. I don't think it's appropriate to bog down the code or core API documentation with loads of comments on how templates work at a basic level. Especially since most of what you've got here truly is clear if considered from the POV of someone who has passing comfort with templating.

I think this will have to be a cooperative endeavour though. You (@Leo) should certainly make sure you are documenting all of the tricks and details that you think are worthwhile, but everyone else looking at this also needs to pipe up on specific things done here that might be worth stuffing in an implementation doc. Particularly people like @chutsu or whoever else (arun?) who come from more of a C/C++ background.

It's just tough to document something through comments that essentially boils down to covering a skill. I don't think putting a few pages of how-to-template style items into comment blocks in the code is an effective approach. I'm not sure what the best approach is, but we do need to be aware of the fact that our overall skill level with templating isn't great and try to account for it.

arunxdas:
@leokoppel @michaelhsmart @StanJBrown I agree that we should be careful with code implementations that are inaccessible to students in the long run.  That said, I'm all for the education of our team, on both the code and math/theory sides.  Though we tend to focus on the algorithms in the reading course, it may be worthwhile for Leo do give a presentation to the lab on TMP, generic programming, if these are tools we see ourselves using day-to-day (not part of the reading course, but perhaps a lab or separate presentation).  @stevenwaslander may have more to comment on this.

StanJBrown:
I agree with you guys fully, and @leokoppel, to be clear I did not mean to say that I didnt like your code, just that I can see it being very intimidating for others to use. I know when I started here, I would be completely lost.  Also, when people look up the docs available for some of these functions / types, such as **decltype(),** you sometimes get stuff like [this](http://en.cppreference.com/w/cpp/language/decltype),  which if I am honest, are not great for learning. That said, it is certainly worth while using this stuff, but we probably need to make some of our own documentation so that people can learn and understand what they are seeing when they look at your code.

michaelhsmart:
@arunxdas A talk might be good. Though it might be a while before he comes up in the talk rotation again. He just presented on testing during this term.

Also, I think everyone following the code of interest should have to contribute to the discussion by highlighting the items that need clarification from their perspective. I don't think the prospective teacher will be able to accurately guesstimate where we're all at with enough accuracy to close out whatever questions we might come up with. Plus if people bring up the questions they have, there's a good chance I can answer a lot of them and offload some of the work in that respect (just DM me on slack with a link to the discussion you want me to chime in on).

Lastly, I want to elaborate on usage of generic programming. I am very much against going too far with generic programming for two reasons. First, it distances us significantly from the general level of incoming students, as Arun alluded to. Second, type inference bugs can be truly nasty and we don't have enough people in the lab with the proficiency to hunt them if they become too common. I haven't had to comment on this topic so far since it's just been Leo using them and he's been good about not going nuts with them at all.

A snippet I mostly agree with from Google's C++ Style Guide:

> Think twice before using template metaprogramming or other complicated template techniques; think about whether the average member of your team will be able to understand your code well enough to maintain it after you switch to another project, or whether a non-C++ programmer or someone casually browsing the code base will be able to understand the error messages or trace the flow of a function they want to call. 

This is a unique challenge for a lab of (mostly) grad students as turn-around is so extreme. In 2 years, most of you (and possibly myself) will be gone. Don't just look at the team as-is when considering the above, but factor in the uncertainty about an expanding/changing team as well. With this uncertainty, how confident are you that there will be someone around who will be able to maintain what you've written?

Again, this specific code would be a case where this level of templating is very appropriate (how else would you even do this?). We just need to make sure we leave enough breadcrumbs for future first-term lab members to not quit in tears if they have to maintain it.

My dream would be that we have a somewhat curated list of template FAQs/Guidelines from around the internet that either provide an "Oh yes, I remember that" answer to people's questions or can at least refer them to something more substantive. Or Leo can reply with a perfect "templates 101" link that will completely resolve our questions.

leokoppel:
Template resources:
- [C++ FAQ: Templates](https://isocpp.org/wiki/faq/templates#overview-templates)
- [Where and why do I have to put the “template” and “typename” keywords?](http://stackoverflow.com/q/610245/431033)
- _Effective C++_, Chapter 7: Templates and Generic Programming 

`decltype` is a keyword introduced in C++11: 
- [C++ auto and decltype Explained](http://thbecker.net/articles/auto_and_decltype/section_01.html)
- [C++ FAQ: decltype](https://isocpp.org/wiki/faq/cpp11-language#decltype)
- _Effective Modern C++_, Item 3

leokoppel:
@chutsu Travis failing during ceres build

chutsu:
Can you try and pull a specific git tag in the ceres install script?

leokoppel:
@chutsu #46

wave_containers/include/wave/containers/measurement_container.hpp
```diff
@@ -0,0 +1,313 @@
...
+    // First, define which members of the Measurement object are used as keys
+    // Specify that the time key corresponds to the time_point member
+    struct time_key : member<T, TimeType, &T::time_point> {};
+    // Specify that the sensor key corresponds to the sensor_id member
+    struct sensor_key : member<T, decltype(T::sensor_id), &T::sensor_id> {};
+
+    // Define a composite key which combines the above two keys
+    // This lets us search elements sorted by sensor id first, then time
+    struct sensor_and_time_key : composite_key<T, sensor_key, time_key> {};
+
```
arunxdas: Will it be easy to add additional keys for things like landmark ids, once we start using these to organize camera data?

leokoppel: Yes, coming soon.

wave_containers/include/wave/containers/measurement_container.hpp
```diff
@@ -0,0 +1,313 @@
...
+        return 0;
+    }
+    composite.erase(it);
+    return 1;
+}
+
+template<typename T>
+const typename MeasurementContainer<T>::ValueType
+MeasurementContainer<T>::get(const TimeType &t, const SensorIdType &s) const {
+    const auto &composite = this->composite();
```
chutsu: Avoiding the `this` keyword eh?

wave_containers/include/wave/containers/measurement.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+using TimeType = std::chrono::steady_clock::time_point;
+
+/** Generic measurement storage type designed to be used with
+ * MeasurementContainer. */
+template<typename T, typename S>
+struct Measurement {
+    TimeType time_point;
+    S sensor_id;
+    T value;
```
leokoppel: Yes. You want the sensor_id type to be an enum class such as in the test:

```cpp
enum class SomeSensors { S1, S2, S3 };
```

That way you can't ask for Lidar #2 when you only have one Lidar.

wave_containers/include/wave/containers/measurement_container.hpp
```diff
@@ -0,0 +1,313 @@
...
+    struct time_index {};
+    struct sensor_index {};
+    struct composite_index {};
+
+    // Define an index for each key. Each index will be accessible via its tag
+    struct indices : indexed_by<
+            ordered_non_unique<tag<time_index>, time_key>,
+            ordered_non_unique<tag<sensor_index>, sensor_key>,
+            ordered_unique<tag<composite_index>, sensor_and_time_key>> {
+    };
```
leokoppel: no, will do

wave_containers/include/wave/containers/measurement_container.hpp
```diff
@@ -0,0 +1,313 @@
...
+     * Note this does *not* correspond to a typical container's value_type. */
+    using ValueType = decltype(MeasurementType::value);
+    /** Alias for template parameter giving the type of the sensor id */
+    using SensorIdType = decltype(MeasurementType::sensor_id);
+
+    using iterator = typename
+    internal::measurement_container<T>::composite_type::iterator;
+    using const_iterator = typename
+    internal::measurement_container<T>::composite_type::const_iterator;
+    using size_type = std::size_t;
```
leokoppel: It is public and can be used from outside the class, just like `std::vector<T>::size_type` (which you should technically use for vector operations instead of `std::size_t`, if you're not using `auto`).

(That it is set to `std::size_t` is a shortcut, yes, but the point is clients don't have to take shortcuts)

wave_containers/include/wave/containers/measurement_container.hpp
leokoppel: Yes. As I suggested in your PR, `const auto &` makes it clear that's what I'm doing.

wave_containers/include/wave/containers/measurement.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+/** Perform linear interpolation (or extrapolation) between two measurements.
+ *
+ * This is the default interpolation function for generic Measurement types.
+ * For measurements requiring non-linear interpolation, you may provide your own
+ * specialization.
+ */
+template<typename T>
+decltype(T::value) interpolate(const T &m1, const T &m2, const TimeType &t) {
+    auto w2 = 1.0 * (t - m1.time_point) / (m2.time_point - m1.time_point);
```
leokoppel: Same as if you type `auto foo = 1.0 / 0`. The client should not do that.

chutsu: fair enough

wave_containers/include/wave/containers/measurement.hpp
```diff
@@ -0,0 +1,38 @@
+#ifndef WAVE_CONTAINERS_MEASUREMENT_HPP
+#define WAVE_CONTAINERS_MEASUREMENT_HPP
+
+#include <chrono>
+
+namespace wave {
+
+using TimeType = std::chrono::steady_clock::time_point;
```
leokoppel: Not a fixed decision, just time_point is more convenient to use (it has math and comparison operators already defined) and explicitly represents a time (as opposed to a duration). It could change it or make it flexible like the sensor and value types.

chutsu: sure thing

chutsu:
@leokoppel when you're ready can you merge this in please?

Issue #40, opened 2017-04-05
VTK install script does not work in Ubuntu 16.04 


chutsu:
The current vtk install script does not work in ubunt 16.04, please modify it a little so that it recognizes what version the user is using. You use the [`ceres_install.bash` ](https://github.com/wavelab/libwave/blob/master/scripts/install/ceres_install.bash) as an example. Note that we can install  libpcl 1.7 in Ubuntu 16.04 via `apt-get` is that good enough?

leokoppel:
What is the VTK install script? Where do we use VTK?

chutsu:
PCL

Issue #39, opened 2017-04-05
VTK compiler warnings


chutsu:
We are currently getting VTK compiler warnings, we need to fix this:

Example:
```
...
-- The imported target "vtkRenderingJava" references the file
   "/usr/lib/jni/libvtkRenderingJava.so.5.8.0"
but this file does not exist.  Possible reasons include:
* The file was deleted, renamed, or moved to another location.
* An install or uninstall procedure did not complete successfully.
* The installation package was faulty and contained
   "/usr/lib/vtk-5.8/VTKTargets.cmake"
but not all the files it references.

-- The imported target "vtkRenderingPythonTkWidgets" references the file
   "/usr/lib/libvtkRenderingPythonTkWidgets.so.5.8.0"
but this file does not exist.  Possible reasons include:
* The file was deleted, renamed, or moved to another location.
* An install or uninstall procedure did not complete successfully.
* The installation package was faulty and contained
   "/usr/lib/vtk-5.8/VTKTargets.cmake"
but not all the files it references.
...
```

Source: https://travis-ci.org/wavelab/libwave/builds/217261477

chutsu:
Also fix this please: 

```
/home/chutsu/Dropbox/proj/libwave/wave_matching/tests/icp_tests.cpp:33:26: warning: deleting object of polymorphic class type ‘wave::ICPMatcher’ which has non-virtual destructor might cause undefined behaviour [-Wdelete-non-virtual-dtor]
             delete this->matcher;
```

PR #43, opened 2017-04-07
Fix PCL Compiler Warnings on Ubuntu 14.04

Jebediah:
This is to address issues #39 and #40. The vtk compiler warnings are because the version of pcl in the ubuntu repositories is 1.7.2 and looks for vtk 5.8 while the version of vtk in the ubuntu repositories is a mix of 5.8, 5.10, and 6. The components being complained about don't seem to be available in version 5.8. Fortunately vtk stands for visual toolkit, and we aren't using these features so these warnings are safe to ignore.

The pcl install script now changes for trusty or xenial.

Finally another warning about virtual destructors was resolved.

### DIFF
--- a/wave_matching/src/icp.cpp
+++ b/wave_matching/src/icp.cpp
@@ -31,6 +31,18 @@ ICPMatcher::ICPMatcher(float res) {
     this->icp.setEuclideanFitnessEpsilon(fit_eps);
 }
 
+ICPMatcher::~ICPMatcher() {
+    if (this->ref) {
+        this->ref.reset();
+    }
+    if (this->target) {
+        this->target.reset();
+    }
+    if (this->final) {
+        this->final.reset();
+    }
+}
+
 void ICPMatcher::setRef(const PCLPointCloud &ref) {
     if (this->resolution > 0) {
         this->filter.setInputCloud(ref);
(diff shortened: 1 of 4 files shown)


leokoppel:
> Fortunately vtk stands for visual toolkit, and we aren't using these features so these warnings are safe to ignore.

So this PR does not actually remove the warnings? (I see travis is still printing them)

Also, why do we install `libvtk`-related packages if we don't use them?

Jebediah:
Yes it doesn't remove them.
vtk is a PCL dependency.

leokoppel:
@Jebediah 

>  Yes it doesn't remove them.

This PR is called "Fix PCL Compiler Warnings on Ubuntu 14.04" so you can understand the confusion - what does this PR actually do then?

> vtk is a PCL dependency.

It is a dependency only of the `visualization` module.

wave_matching/src/icp.cpp
```diff
@@ -31,6 +31,18 @@ ICPMatcher::ICPMatcher(float res) {
...
+    if (this->ref) {
+        this->ref.reset();
+    }
+    if (this->target) {
+        this->target.reset();
+    }
+    if (this->final) {
+        this->final.reset();
+    }
+}
```
chutsu: Are these resets actally needed? won't they reset themselves when they go out of scope? Just a question.

chutsu:
The title is my bad it was originally "Fix 39".

wave_matching/src/icp.cpp
Jebediah: wasn't sure. decided to play it safe

Issue #14, opened 2017-03-17
Refactor Travis process


chutsu:
Currently the Travis CI process takes too long (~20minutes), we could like to utilize Travis caching in-order to speed up travis builds, culpret for the long build is OpenCV.

leokoppel:
Builds seem to be taking longer than expected; reopening.

leokoppel:
See #76

PR #46, opened 2017-04-13
Install script: use ceres release 1.12.0

leokoppel:
Currently we are pulling and building master, which sometimes causes issues.

The `CERES_VERSION` variable will have to be updated when we want to install a
newer release. (It is possible to check out the latest tag, but that includes
release candidate tags, so would need more logic, etc)

### DIFF
--- a/scripts/install/ceres_install.bash
+++ b/scripts/install/ceres_install.bash
@@ -2,6 +2,7 @@
 set -e  # exit on first error
 UBUNTU_VERSION=`lsb_release --release | cut -f2`
 SRC_PREFIX_PATH="/usr/local/src/"
+CERES_VERSION=1.12.0
 
 
 install_dependencies() {
@@ -27,6 +28,7 @@ install_ceres_solver() {
 
     # go into ceres-solver repo and prepare for build
     cd ceres-solver
+    git checkout tags/${CERES_VERSION}
     mkdir build
     cd build
     cmake ..


scripts/install/ceres_install.bash
```diff
@@ -2,6 +2,7 @@
 set -e  # exit on first error
 UBUNTU_VERSION=`lsb_release --release | cut -f2`
 SRC_PREFIX_PATH="/usr/local/src/"
```
yuzhangbit: Is there any special reason to put source codes in `/usr/local/src`?

yuzhangbit: As the source codes is kept in `/usr/local/src` folder, this script will fail at the second time you install ceres(or update ceres) .

leokoppel: No, just a convention / Chris' preference. Not in scope of this PR.

leokoppel: I guess you mean the line `git clone` will fail? Yes.. I will fix it for issue #14.

chutsu: @yuzhangbit is there a reason why `/usr/local/src` isn't a good idea, I was merely following the [FHS standard](https://unix.stackexchange.com/questions/10814/what-is-the-usr-local-src-folder-meant-for)? I've also thought about creating our own `/opt/wavelab` I'm open to options / ideas

yuzhangbit: No. I am just curious.  Good to know the standard!  As all the 3rd-party libraries in `/usr/local/src` have been installed and the `build` folders are not needed, is it a good practice to delete the `build` folder to save space?   The build folders can be really big after several libraries installed.

chutsu: Indeed they can get big but they also save time during compilation, so it is 50/50 decision to remove it or not, I opted not to by default because space is rarely an issue for modern systems, of course there are exceptions.

leokoppel: Yeah I have come around that /usr/local/src is a correct place to put source code. Still not sure if this script should do it. In autonomoose we just put the builds into `tmp`. The reasoning being someone running the "install_dependencies" script just wants dependencies installed, and didn't ask for or expect source code in particular. Indeed most of the scripts just run `apt-get`, and we could change this ceres script to use `apt-get` tomorrow.

.. I agree with your comment elsewhere that "we should maybe create Ubuntu packages and use apt-get install"

leokoppel:
Closing as this change was included in #48

PR #48, opened 2017-04-14
Add cache to travis builds

### LINKED ISSUE #14: Refactor Travis process

leokoppel:
A number of changes to improve the dependency install step. Resolves #14

### DIFF
--- a/scripts/install/ceres_install.bash
+++ b/scripts/install/ceres_install.bash
@@ -39,11 +39,10 @@ install_ceres_solver() {
     cmake ..
 
     # compile and install
-    make
+    make -j$(nproc)
     make test
     make install
 }
-
 
 # MAIN
 if [ $UBUNTU_VERSION == "16.04" ]; then
--- a/scripts/install/ceres_install.bash
+++ b/scripts/install/ceres_install.bash
@@ -21,15 +21,20 @@ install_suitesparse_fix() {
 }
 
 install_ceres_solver() {
-    # clone ceres solver
     mkdir -p $SRC_PREFIX_PATH
     cd $SRC_PREFIX_PATH
-    git clone https://github.com/ceres-solver/ceres-solver.git
+
+    # clone ceres-solver if directory does not already exist, or pull
+    if [ ! -d ceres-solver ]; then
+       git clone https://github.com/ceres-solver/ceres-solver.git
+    else
+       git -C ceres-solver pull
+    fi
 
     # go into ceres-solver repo and prepare for build
     cd ceres-solver
     git checkout tags/${CERES_VERSION}
-    mkdir build
+    mkdir -p build
     cd build
     cmake ..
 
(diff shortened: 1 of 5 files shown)


scripts/install/ceres_install.bash
```diff
@@ -20,24 +21,30 @@ install_suitesparse_fix() {
...
-    mkdir build
+    git checkout tags/${CERES_VERSION}
+    mkdir -p build
     cd build
     cmake ..
 
     # compile and install
-    make
-    make test
+    make -j8
```
yuzhangbit: `make -j$(nproc)` ?

scripts/install/ceres_install.bash
```diff
@@ -20,24 +21,30 @@ install_suitesparse_fix() {
...
-    # clone ceres solver
     mkdir -p $SRC_PREFIX_PATH
     cd $SRC_PREFIX_PATH
-    git clone https://github.com/ceres-solver/ceres-solver.git
+
+    # clone ceres-solver if directory does not already exist, or pull
+    if [ ! -d ceres-solver ]; then
+       git clone https://github.com/ceres-solver/ceres-solver.git
+    else
+       git -C ceres-solver pull
```
yuzhangbit: This line won't work the second time as you check out a tag version(you are not in any branch) later on.

chutsu: In any case I don't believe you need to pull, because we are checking out a specific tag.

scripts/install/install_deps.bash
```diff
@@ -5,10 +5,20 @@ sudo apt-get clean > /dev/null
...
-    if [[ $? -ne 0 ]]; then
+    if [[ $- == *i* ]]; then # If shell is interactive
+    # Turn off buffering in awk (so dots print right away)
+        local awk_arg="-W interactive"
+    fi
+
+    # Run the install script. Print a dot for every 10 lines of output and save
+    # the full output to a log file
+    sudo bash $2 |& tee install.log \
+        |& awk ${awk_arg} 'NR%10==1 { printf "."} END{ print ""}'
```
chutsu: Just a comment, but this is very clever indeed, my guess is that you're trying to print dots so that we don't need the `travis_wait` in the `travis.yaml` and kill off our build because it didn't print anything for 10 minutes.

scripts/install/ceres_install.bash
chutsu: In anycase I'm not certain you actually need to pull, since you are checking out a specific tag.

leokoppel: You're right. I will change to `fetch`. Thinking of the case where we change the version and the new tags aren't downloaded.

scripts/install/install_deps.bash
leokoppel: Yes, described in commit message.

leokoppel:
> we should maybe standardize the way packages are installing in another issue, 

Good idea. Though if we put too much work into it we'll have a fully fledged package manager..

chutsu:
I honestly think that we should maybe create Ubuntu packages and use `apt-get install` (shotgun not me)

PR #54, opened 2017-04-20
Add exception handling to rotation class.

arunxdas:
This PR adds invalid_argument exceptions and unit tests for input arguments of the rotation class member
functions.  Specifically, it checks for:

* Nan input elements
* Inf input elements
* Valid rotation matrix


wave_geometry/include/wave/geometry/exception_helpers.hpp
```diff
@@ -0,0 +1,49 @@
...
+void checkValidRotation(const Mat3 &R) {
+    // fltcmp returns 0 on successful comparision.
+    if (!Rotation::isValidRotationMatrix(R)) {
+        throw std::invalid_argument(
+          "Matrix input is not a valid rotation matrix.");
+    }
+}
+}
+
+#endif
```
Jebediah: needs empty line

leokoppel:
These checks could be useful. Data should be checked for validity whenever it comes from external sources (user input, reading a file, etc).

Whether "external sources" includes "programmer input from outside the library" is up for debate. (I would expect a library to, at most, `assert` this kind of input, but that's up for debate.)

Checking on entirely internal operations seems superfluous. Where do you expect NaNs to come from before every `rotate` call, for example?

arunxdas:
The main concern is if the input to the function is the result of some other intermediate calculation that could result in an inf or NaN.  For example, some other kinematics or un-project math that doesn't check if the resulting vector is a valid value.  In general, I thought it was good practice to check function input for validity, but if these seem superfluous, we can remove.  Or at least keep a subset of them where it makes sense.

leokoppel:
Oh my mistake - `rotate` isn't an "entirely internal operation" because we haven't wrapped the Vector type. It is often good practice but usually not done in a math library (see discussion [here](https://softwareengineering.stackexchange.com/questions/64926/should-a-method-validate-its-parameters), for example).

In my opinion, if some client is passing a vector to the rotate function, it is their job to check that it's valid - that includes not being NaN, as well as simply being "correct" according to their application. Most  possible "incorrect" inputs are impossible for us to check. (though explicit reference frames could help with some common mistakes).

It's okay to keep the checks for now, and possibly remove them when performance is reviewed.

wave_geometry/src/rotation.cpp
```diff
@@ -66,17 +87,37 @@ Rotation &Rotation::setFromEulerXYZ(const Vec3 &input_vector) {
...
 Rotation &Rotation::setFromAngleAxis(const double angle_magnitude,
                                      const Vec3 &rotation_axis) {
+    // Check if the input is finite, throw error otherwise
+    checkArrayFinite(rotation_axis);
+    checkScalarFinite(angle_magnitude);
+
+    // Check if the input rotation axis is normalized,
+    // throw error otherwise.
+
+    checkVectorNormalized(rotation_axis);
```
leokoppel: Note: Kindr does this:

>  In debug mode, an assertion is thrown if the rotation vector has not unit length.

wave_geometry/include/wave/geometry/exception_helpers.hpp
```diff
@@ -0,0 +1,49 @@
...
+void checkVectorNormalized(const Eigen::MatrixBase<MatrixType> &input_vector) {
+    // fltcmp returns 0 on successful comparision.
+    if (fltcmp(input_vector.norm(), 1)) {
+        throw std::invalid_argument("Vector input is not normalized.");
+    }
+}
+
+// Function that checks to see if the input matrix is a valid rotation.
+void checkValidRotation(const Mat3 &R) {
+    // fltcmp returns 0 on successful comparision.
```
leokoppel: wrong comment

wave_geometry/include/wave/geometry/rotation.hpp
```diff
@@ -57,6 +57,9 @@ class Rotation {
     // Operator overloads.
     Rotation operator*(const Rotation &R) const;
     friend std::ostream &operator<<(std::ostream &stream, const Rotation &R);
+
+    // Static functions.
+    static bool isValidRotationMatrix(const Mat3 &input_matrix);
```
leokoppel: This should not be a member function, since it does not use any non-public members of `Rotation` (or _any_ members of `Rotation`, for that matter).

(see http://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197)

michaelhsmart:
I personally like keeping pre-condition checks in for the purposes of easy debugging etc., and then paring them down afterwards when you are working on optimization.

For assert vs other methods, since asserts are guaranteed to be fatal if they are enabled, I prefer for them to not be used.

If there's any possibility of an end user triggering the scenario for an assert, then don't use an assert. If the scenario can only be due to internal error and you can guarantee that and can then test to catch all such errors, then asserts are fine. The problem is that depending on the problem it can be extremely difficult (if not impossible) to achieve that which is why we've said no to asserts in general.

leokoppel:
I think the use of assert in Kindr and Eigen is appropriate "for the purposes of easy debugging etc., and then paring them down afterwards when you are working on optimization".

> If there's any possibility of an end user triggering the scenario for an assert

-agree, but library calls are not "end user" input

wave_geometry/include/wave/geometry/exception_helpers.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include "wave/utils/math.hpp"
+
+namespace wave {
+
+
+// This function checks to make sure all elements of the input
+// Eigen array have finite values (not Nan or +/- Inf), as these
+// values are invalid for rotation geometry.
+template <typename MatrixType>
+void checkArrayFinite(const Eigen::MatrixBase<MatrixType> &mat) {
```
chutsu: Minor naming thing, but I don't know if this should be named `checkMatVecFinite` instead of array? or even specific like `checkVecFinite` and `checkMatFinite`? last thing do you think maybe C style functions like this should be snake case (i.e. all lower, separated by underscores: `this_is_a_snake`)? to be able to quickly differentiate between class methods vs functions. Food for thought.

wave_geometry/include/wave/geometry/exception_helpers.hpp
```diff
@@ -0,0 +1,49 @@
...
+    if (!std::isfinite(input_scalar)) {
+        throw std::invalid_argument("Scalar input is non-finite.");
+    }
+}
+
+// Function that checks to see if the input vector is normalized.
+template <typename MatrixType>
+void checkVectorNormalized(const Eigen::MatrixBase<MatrixType> &input_vector) {
+    // fltcmp returns 0 on successful comparision.
+    if (fltcmp(input_vector.norm(), 1)) {
```
chutsu: Not quite ideal:
1. I think you should feed in a specific epsilon (error threshold) yourself, 
2. `fltcmp(input_vector.norm(), 1.0)` if I didn't write this myself I wouldn't know that a return of `0` means "equal", I would have just added `== 0` just to be clear.

Example:
```
void checkVectorNormalized(const Eigen::MatrixBase<MatrixType> &input_vector, double eps=0.0001) {
  // fltcmp returns 0 on if equal
  if (fltcmp(input_vector.norm(), 1.0, eps) != 0) {
```

wave_geometry/src/rotation.cpp
```diff
@@ -26,16 +35,25 @@ void Rotation::invert() {
...
+    checkArrayFinite(input_vector);
+
     return this->rotation_object.rotate(input_vector);
 }
 
 Vec3 Rotation::inverseRotate(const Vec3 &input_vector) const {
+    // Check if the input is finite, throw error otherwise
+    checkArrayFinite(input_vector);
+
     return this->rotation_object.inverseRotate(input_vector);
```
chutsu: What does Kindr do if this fails?

wave_geometry/src/rotation.cpp
chutsu: I think having our own `checkVectorNormalized` is fine for now, I mean the ultimate goal is to eventually have our own version of Kindr right?

wave_geometry/src/rotation.cpp
```diff
@@ -66,17 +87,37 @@ Rotation &Rotation::setFromEulerXYZ(const Vec3 &input_vector) {
...
+
+    checkVectorNormalized(rotation_axis);
+
     kindr::AngleAxisD angle_axis(angle_magnitude, rotation_axis);
     this->rotation_object = angle_axis;
     return *this;
 }
 
 Rotation &Rotation::setFromExpMap(const Vec3 &se3_vector) {
+    // Check if the input is finite, throw error otherwise
```
chutsu: superfluous comment

wave_geometry/src/rotation.cpp
```diff
@@ -66,17 +87,37 @@ Rotation &Rotation::setFromEulerXYZ(const Vec3 &input_vector) {
...
+    checkArrayFinite(se3_vector);
+
     this->rotation_object = this->rotation_object.exponentialMap(se3_vector);
     return *this;
 }
 
 Rotation &Rotation::setFromMatrix(const Mat3 &input_matrix) {
+    // Check if the input is finite, throw error otherwise
```
chutsu: superfluous comment

wave_geometry/src/rotation.cpp
```diff
@@ -66,17 +87,37 @@ Rotation &Rotation::setFromEulerXYZ(const Vec3 &input_vector) {
...
+    checkArrayFinite(input_matrix);
+
+    // Check if the input is a valid rotation matrix,
+    // throw error otherwise.
```
chutsu: superfluous comment

wave_geometry/src/rotation.cpp
```diff
@@ -86,12 +127,19 @@ Vec3 Rotation::logMap() const {
 }
 
 Rotation &Rotation::manifoldPlus(const Vec3 &omega) {
+    // Check if the input is finite, throw error otherwise
```
chutsu: superfluous comment

wave_geometry/src/rotation.cpp
```diff
@@ -86,12 +127,19 @@ Vec3 Rotation::logMap() const {
...
+    checkArrayFinite(omega);
+
     this->rotation_object = this->rotation_object.boxPlus(omega);
     return *this;
 }
 
 bool Rotation::isNear(const Rotation &R,
-                      const double comparison_threshold) const {
+                      const double comparison_threshold = 1e-6) const {
+    // Check if the input is finite, throw error otherwise
```
chutsu: superfluous comment

wave_geometry/src/rotation.cpp
```diff
@@ -101,6 +149,9 @@ Mat3 Rotation::toRotationMatrix() const {
 }
 
 Rotation Rotation::operator*(const Rotation &R) const {
+    // Check if the input is finite, throw error otherwise
```
chutsu: superfluous comment

wave_geometry/include/wave/geometry/exception_helpers.hpp
leokoppel: `checkMatrixFinite` seems good. We're already using `this->` to clearly indicate methods right? What is "C style" about this function?

chutsu: I was merely refering to the naming conventions for functions vs class methods, do we want to clearly differentiate it with `snake_case` vs `camelCase`

leokoppel: I think we already differentiate class methods using the `this->` keyword. Functions can have `camelCase`.

wave_geometry/src/rotation.cpp
arunxdas: You mean if the inverseRotate fails?

chutsu: Yea, is it throwing signs exception or kindr's own custom one?

chutsu:
Is this ready to be merged in?

arunxdas:
I need to do some quick clean up, should be able to get to it this week.
Mainly just fixing comments

On Mon, May 15, 2017 at 12:15 PM, Christopher Choi <notifications@github.com

chutsu:
@arunxdas **cough**, **cough**

leokoppel:
@arunxdas @chutsu I will fix build errors here. Please review my other PRs

PR #56, opened 2017-04-22
General code clean up with wave_geometry and wave_matching

chutsu:
- Make CMake files more consistent across modules
- Minor test code formatting in wave_matching
- Add `wave_geometry_test` to travis as part of CI-build
- Remove / reorganize headers in wave_matching test code
- Remove duplicated `wave_matching/tests/testscan.pcd` (its already in `wave_matching/tests/data/testscan.pcd`
- Rename g_icp.cpp to gicp.cpp and its header
- Add copy of wave_matching configs into test folder
- Unify all test runner code


wave_matching/include/wave/matching/matcher.hpp
```diff
@@ -47,9 +47,9 @@ class Matcher {
 
  protected:
     float resolution;
-    Eigen::Affine3d result;
-    Eigen::MatrixXd information;
+    Affine3 result;
+    MatX information;
```
Jebediah: Are these our datatypes?

Jebediah: nevermind found the typedefs

PR #57, opened 2017-04-24
Draft factor graph implementation

chutsu:
The idea of this PR is to implement a basic factor graph implementation. Docs to follow shortly...


wave_containers/tests/containers/measurement_test.cpp
```diff
@@ -113,7 +113,6 @@ TEST(Utils_measurement, get_interpolated_other_sensors) {
     auto t2 = t1 + std::chrono::seconds(10);
     auto tmid = t1 + std::chrono::seconds(5);
 
-    auto v1 = 3.5, v2 = 8.0;
```
chutsu: @leokoppel was this needed?

wave_containers/tests/containers/measurement_test.cpp
```diff
@@ -132,7 +131,6 @@ TEST(Utils_measurement, iterators) {
 
     for (auto i = 0; i < 5; ++i) {
         auto now = std::chrono::steady_clock::now();
-        auto value_in = i * 1.1;
```
chutsu: @leokoppel was this needed?

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+namespace wave {
+
+class FactorGraph {
+ public:
+    std::multimap<std::shared_ptr<Variable>, std::shared_ptr<Factor>> graph;
+    std::set<std::shared_ptr<Variable>> variables;
+    std::vector<std::shared_ptr<Factor>> factors;
+
+    FactorGraph() : graph{}, variables{}, factors{} {}
```
leokoppel: Unnecessary, the default constructors are already called

wave_optimization/include/wave/optimization/factor_graph/factor.hpp
```diff
@@ -0,0 +1,69 @@
...
+        return oss.str();
+    }
+
+    virtual void print() {
+        std::cout << this->toString() << std::endl;
+    }
+};
+
+template <typename T>
+struct UnaryFactor : Factor {
```
leokoppel: So we shouldn't separate function bodies and declarations? (I asked this question in my PR and didn't get feedback)

wave_optimization/include/wave/optimization/factor_graph/factor.hpp
```diff
@@ -0,0 +1,69 @@
...
+        std::ostringstream oss;
+
+        switch (this->arity) {
+            case 1:
+                oss << "[factor_type: unary, ";
+                oss << "at: " << this->variables[0]->toString();
+                oss << "]";
+                break;
+            case 2:
+                oss << "[factor_type: binary, ";
```
leokoppel: A bit confusing - if the base class already knows everything about arity, what's the point of having separate `UnaryFactor` and `BinaryFactor` classes? E.g., why is this method `virtual`, but has this conditional inside it instead of being overridden in each derived class?

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+    template <typename V, typename T>
+    int addUnaryFactor(size_t at, const T &z) {
+        // store variable and factor
+        auto variable = std::make_shared<V>(at);
```
leokoppel: Nice `auto` and `make_shared` :+1:

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+        this->variables.insert(variable);
+
+        auto factor = std::make_shared<UnaryFactor<T>>(variable, z);
+        this->factors.push_back(factor);
+
+        // store key-factor pair
+        auto value = std::make_pair(variable, factor);
+        this->graph.insert(value);
+
+        return 0;
```
leokoppel: What does this mean?

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    std::string toString() {
+        std::ostringstream oss;
+        int index = 0;
+
+        for (auto f : this->factors) {
+            oss << "f" << index << " -- ";
+            oss << f->toString() << std::endl;
```
leokoppel: Implement `operator<<` for the factors, programmers expect to be able to use `oss << f << std::endl;`

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef WAVE_OPTIMIZATION_FACTOR_GRAPH_VARIABLE_HPP
+#define WAVE_OPTIMIZATION_FACTOR_GRAPH_VARIABLE_HPP
+
+#include "wave/utils/utils.hpp"
+
+namespace wave {
+
+struct Variable {
+    size_t id;
```
leokoppel: Only use `size_t` for sizes.. I actually thought all the `size_t` arguments in this PR were sizes, when they referred to "id"

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    VecX &getData() {
+        return data;
+    }
+};
+
+struct PoseVar : Variable {
+    PoseVar(size_t id) : Variable{id, 6} {
+        this->data.fill(0.0);
```
leokoppel: It is possible to zero-initialize inside the base constructor - because you do give the base constructor a size.

wave_optimization/tests/factor_graph/factor_graph_test.cpp
```diff
@@ -0,0 +1,39 @@
+#include "wave/wave_test.hpp"
+#include "wave/optimization/factor_graph/factor_graph.hpp"
+
+
+namespace wave {
+
+TEST(FactorGraph, addUnaryFactor) {
+    FactorGraph graph;
+    graph.addUnaryFactor<PoseVar, double>(1, 2.0);
```
Jebediah: I don't understand the change. Are the templates hidden somehow?

leokoppel: The templates are still there, and are not really the issue.

#### What interface to use?

The change I am suggesting is an interface change from

```cpp
void addUnaryFactor(int, double);
```
to something like:
```cpp
void addUnaryFactor(PoseVar, UnaryFactor);
```

Which IMO is considerably more clear. The above interface decision does not involve templates.

#### But what about the templates?

Of course, the above signatures are what the _template function_ looks like after is generated from a _function template_ that looks like this:
```cpp
template<typename V, typename T>
void addUnaryFactor(int, const &T);
```
or
```cpp
template<typename V, typename T>
void addUnaryFactor(V, T);
```

#### How do templates affect the interface?

Since we _are_ using templates, there is a further difference between the two options: in the first, the compiler cannot deduce the template parameter `V` from a call to the function. That is why the call must [explicitly specify which version of the template you want](https://isocpp.org/wiki/faq/templates#fn-templates-explicit-calls):

```cpp
addUnaryFactor<PoseVar, double>(1, 2.0);
```
vs.
```cpp
addFactor(PoseVar{1}, UnaryFactor{2.0});
```

IMO the second option is again more clear.

Note the second call _could_ be written as 
`addFactor<PoseVar, UnaryFactor>(PoseVar{1}, UnaryFactor{2.0});`
but it is not necessary.

###Template resources
- [C++ FAQ: Templates](https://isocpp.org/wiki/faq/templates#overview-templates)
- [Where and why do I have to put the “template” and “typename” keywords?](http://stackoverflow.com/q/610245/431033)
- _Effective C++_, Chapter 7: Templates and Generic Programming

Jebediah: I see. Interesting.

chutsu: I am also a proponent of non-template interface, the add function can be generalized to just 

    add(PoseVar{1}, UnaryFactor{2.0})

instead of `addUnaryFunction()` right?

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
chutsu: `at` is an id of a factor graph variable, the choice of `at` can be anything, the most obvious choice is an integer number which we can increment easily by adding 1. Think of it as an id for every pose or landmark we are optimising.

The reason why its `size_t` is because it has a high bound, an `int` is represented by 4 bytes, and has a range of –2,147,483,648 to 2,147,483,647, where as `size_t` can store the maximum size of a theoretically possible object of any type.

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
chutsu: I will typedef it, I don't think I like using `using` for this purpose though.

leokoppel: http://stackoverflow.com/questions/10747810/what-is-the-difference-between-typedef-and-using-in-c11

See _Effective Modern C++_ Item 9: Prefer alias declarations to typedefs

I prefer `using` even where it gives no extra functionality because
- It looks just like an assignment. `typedef` looks "backwards".
- since `using` is necessary in some cases, it is more consistent to use it everywhere.

>  I think using using is a tad more confusing than using typedef to non-modern C++ developers.

Sure, it is confusing the first time you see it, just like the first time a new C programmer sees `typedef`. (probably less so, because of the backwardness issue).

I'm not opposed to you writing `typedef` if you want.

wave_optimization/include/wave/optimization/factor_graph/factor.hpp
arunxdas: instead of having an overloaded Factor() for each type (unary, binary, etc), is it easier to have the Factor function take in a list of keys for the optimization variables involved in the factor?

wave_optimization/tests/factor_graph/factor_graph_test.cpp
arunxdas: How does  UnaryFactor{2.0} instantiate with just one argument though? Doesn't the UnaryFactor need the key as well?

arunxdas: Let's have a quick discussion during Monday's meeting on finalizing this interface.  Also, how will this interface connect to the Ceres solver, @chutsu ?

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
chutsu: Fair enough to I'll be modern

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef WAVE_OPTIMIZATION_FACTOR_GRAPH_VARIABLE_HPP
+#define WAVE_OPTIMIZATION_FACTOR_GRAPH_VARIABLE_HPP
+
+#include "wave/utils/utils.hpp"
+
+namespace wave {
+
+struct Variable {
```
chutsu: Refactored to `FactorVariable`

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+        oss << "[";
+        oss << "id: " << this->id << ", ";
+        oss << "size: " << this->size;
+        oss << "]";
+
+        return oss.str();
+    }
+
+    void print() {
```
chutsu: Done

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    bool operator<(const Variable &other) const {
+        if (this->id < other.id) {
+            return true;
+        }
+        return false;
+    }
+
+    VecX &getData() {
```
chutsu: Removed

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
...
+        return data;
+    }
+};
+
+struct PoseVar : Variable {
+    PoseVar(size_t id) : Variable{id, 6} {
```
chutsu: This goes back to our previous discussion. The factor graph has a storage (be it `std::vector`, etc...) of only the base class `FactorVariable` (Variable - old name), if the factor graph were to operate on this base class `FactorVariable` it will not be able to obtain the derived `PoseVar`'s fixed size vector, because `PoseVar` when added into the factorgraph will have been casted down, if we were to cast a `FactorVariable` back to a `PoseVar` sure,we can have a vector of fixed size of 6, but would you agree this complicates design?

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include "wave/utils/utils.hpp"
+#include "wave/optimization/factor_graph/factor.hpp"
+#include "wave/optimization/factor_graph/variable.hpp"
+
+namespace wave {
+
+class FactorGraph {
+ public:
+    std::multimap<std::shared_ptr<Variable>, std::shared_ptr<Factor>> graph;
```
chutsu: Indeed, for now lets focus more on the interface and come back when we have more of an idea of the complexity of the problem.

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
chutsu: Just being pedantic, is it bad?

wave_optimization/include/wave/optimization/factor_graph/factor.hpp
chutsu: Refactored

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
```diff
@@ -0,0 +1,56 @@
+#ifndef WAVE_OPTIMIZATION_FACTOR_GRAPH_VARIABLE_HPP
+#define WAVE_OPTIMIZATION_FACTOR_GRAPH_VARIABLE_HPP
+
+#include "wave/utils/utils.hpp"
+
+namespace wave {
+
+struct Variable {
+    size_t id;
+    size_t size;
```
chutsu: Done

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
leokoppel: I think I see what you mean.  Yes, if you have a case like this
```cpp
struct Base {};
struct A : public Base {
    ChrisType data;
}
struct DerivedB : public Base {
    LeoType data;
}
std::vector<Base*> v;
v.push_back(new DerivedA());
v.push_back(new DerivedB());
```
You will not be able to directly access `data` through the `Base*`. But `Base` only needs to provide an _interface_, not expose `data` directly. See the chapter on inheritance in _Effective C++_.

There are a variety of possible solutions involving virtual functions, function pointers, and/or templates. 

You never have to "store the variable type in the class". I agree that would be silly.

chutsu: The thing is if I don't have direct access to `data`, will I still be able to grab the address of the data so I can use it in Ceres optimization?

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
leokoppel: It just gets out of hand if you have many members or multiple constructors.

If paranoid you can use a default member initializer:
```cpp
std::vector<std::shared_ptr<Factor>> factors{};
```

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
leokoppel: You can certainly use it in Ceres optimization, not sure if it's a simple as "grab the address".

Keep it this way for now, refactor later?

chutsu: Sure.

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
chutsu: ok

wave_optimization/include/wave/optimization/factor_graph/factor_graph.hpp
chutsu: Done

wave_optimization/include/wave/optimization/factor_graph/variable.hpp
chutsu: Done

wave_optimization/include/wave/optimization/factor_graph/factor.hpp
```diff
@@ -0,0 +1,69 @@
...
+};
+
+template <typename T>
+struct BinaryFactor : Factor {
+    T measurement;
+
+    BinaryFactor(const std::shared_ptr<Variable> &from,
+                 const std::shared_ptr<Variable> &to,
+                 const T &measurement)
+        : Factor{from, to}, measurement{measurement} {}
```
chutsu: well, GTSAM calls it key1, key2, etc... factor of 3 or more is unclear.

PR #60, opened 2017-05-10
Add install target to cmake

chutsu:
Add a "install" target to cmake so that libwave can be installed to `/usr/local/include` and `/usr/local/lib`.

This PR adds the ability to build `libwave` as a shared library or a static library. For development purposes it makes sense to compile a shared library and for deployment a static library is advised. Either way the option is there, by default cmake compiles the former option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ ADD_SUBDIRECTORY(wave_utils)
 # wave_containers: header-only library
 ADD_SUBDIRECTORY(wave_containers)
 
-# libwave
+# libwave library
 SET(LIBWAVE_MODULE_LIBS
     ${CMAKE_BINARY_DIR}/wave_controls/libwave_controls.a
     ${CMAKE_BINARY_DIR}/wave_geometry/libwave_geometry.a
@@ -94,10 +94,7 @@ ELSE ()
 
 ENDIF()
 
-INSTALL(DIRECTORY wave_containers/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_controls/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_geometry/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_kinematics/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_matching/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_optimization/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_utils/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
+# libwave headers
+FOREACH(MODULE ${LIBWAVE_MODULES})
+  INSTALL(DIRECTORY ${MODULE}/include/ DESTINATION include)
+ENDFOREACH()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,22 @@ ADD_SUBDIRECTORY(wave_utils)
 ADD_SUBDIRECTORY(wave_containers)
 
 # libwave
+SET(LIBWAVE_MODULE_LIBS
+    ${CMAKE_BINARY_DIR}/wave_controls/libwave_controls.a
+    ${CMAKE_BINARY_DIR}/wave_geometry/libwave_geometry.a
+    ${CMAKE_BINARY_DIR}/wave_kinematics/libwave_kinematics.a
+    ${CMAKE_BINARY_DIR}/wave_matching/libwave_matching.a
+    ${CMAKE_BINARY_DIR}/wave_optimization/libwave_optimization.a
+    ${CMAKE_BINARY_DIR}/wave_utils/libwave_utils.a)
+
+SET(LIBWAVE_MODULES
+    wave_controls
+    wave_geometry
+    wave_kinematics
+    wave_matching
+    wave_optimization
+    wave_utils)
+
 IF(BUILD_SHARED_LIBS)
     # SHARED LIBRARY
     ADD_CUSTOM_TARGET(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -50,22 +50,58 @@ ADD_SUBDIRECTORY(wave_utils)
...
+            wave_kinematics
+            wave_matching
+            wave_optimization
+            wave_utils
+    )
+    INSTALL(FILES ${CMAKE_BINARY_DIR}/libwave.a DESTINATION /usr/local/lib)
+
+ENDIF()
+INSTALL(DIRECTORY wave_containers/include/ DESTINATION /usr/local/include)
+INSTALL(DIRECTORY wave_controls/include/ DESTINATION /usr/local/include)
```
leokoppel: Also, should these be in the individual modules' CMakeLists? What's the point of having individual CMakeLists if the parent CMakeLists has to know everything about the children?

CMakeLists.txt
```diff
@@ -50,22 +50,58 @@ ADD_SUBDIRECTORY(wave_utils)
...
+            wave_controls
+            wave_geometry
+            wave_kinematics
+            wave_matching
+            wave_optimization
+            wave_utils
+    )
+    INSTALL(FILES ${CMAKE_BINARY_DIR}/libwave.so DESTINATION /usr/local/lib)
+
+ELSE ()
```
chutsu: Done

CMakeLists.txt
```diff
@@ -50,22 +50,58 @@ ADD_SUBDIRECTORY(wave_utils)
...
+            wave_geometry
+            wave_kinematics
+            wave_matching
+            wave_optimization
+            wave_utils
+    )
+    INSTALL(FILES ${CMAKE_BINARY_DIR}/libwave.a DESTINATION /usr/local/lib)
+
+ENDIF()
+INSTALL(DIRECTORY wave_containers/include/ DESTINATION /usr/local/include)
```
chutsu: Done

CMakeLists.txt
chutsu: The idea is flexibility, I would like to declare the whole install instruction all in one place rather than to each of their different modules. Changing this in future can be tiresome. Also if for any particular reason the user or us wishes to not install a specific module we can do so from this point.

CMakeLists.txt
```diff
@@ -50,22 +50,54 @@ ADD_SUBDIRECTORY(wave_utils)
...
+        COMMAND
+            ar -rcT libwave.a ${LIBWAVE_MODULE_LIBS}
+        DEPENDS
+            ${LIBWAVE_MODULES}
+    )
+    INSTALL(FILES ${CMAKE_BINARY_DIR}/libwave.a DESTINATION CMAKE_INSTALL_PREFIX}/lib)
+
+ENDIF()
+
+INSTALL(DIRECTORY wave_containers/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
```
leokoppel: To clarify, use `INSTALL(DIRECTORY wave_containers/include/ DESTINATION include)`

Relative paths are relative to `CMAKE_INSTALL_PREFIX`.

CMakeLists.txt
leokoppel: A project having multiple optional subprojects or modules is a common pattern. Can you find an established open-source project with this pattern and describe what it does?

CMakeLists.txt
```diff
@@ -94,10 +94,7 @@ ELSE ()
...
-INSTALL(DIRECTORY wave_containers/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_controls/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_geometry/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_kinematics/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_matching/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_optimization/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
-INSTALL(DIRECTORY wave_utils/include/ DESTINATION ${CMAKE_INSTALL_PREFIX}/include)
+# libwave headers
+FOREACH(MODULE ${LIBWAVE_MODULES})
+  INSTALL(DIRECTORY ${MODULE}/include/ DESTINATION include)
```
leokoppel: nice

CMakeLists.txt
chutsu: Their requirements are slightly different, the main idea here is I want to create a single library object (be it shared or static) and link against it, rather than a long list of modules to include. If you want the other approach can you do it?

chutsu: And if you do decide to do it the other way, please do it soon. I need this since my gazebo simulation code now relies on this.

leokoppel: Ah OK. Do we still want "a project having multiple optional subprojects or modules"? Note that doesn't say anything about what the build products look like.

If the goal is a single library object, is it necessary to build separate `.a` files for each module?

chutsu: > "Do we still want "a project having multiple optional subprojects or modules"? Note that doesn't say anything about what the build products look like."

We still want to have multiple optional modules, but that will be an issue for later.

> If the goal is a single library object, is it necessary to build separate .a files for each module?

This is tricky, I either compile separate `.a` per module or I have to export the source files from the module to the parent cmake file via `SET(SOURCEFILES main.cpp foo.cpp PARENT_SCOPE)` either approaches will work, the downside is if one copies one of the libwave directly the developer will now have to edit the module level cmake to remove the `PARENT_SCOPE` keyword. 

OR! We just take care of it from the parent cmake file, either way works I preferred the latter.

leokoppel: > And if you do decide to do it the other way, please do it soon. I need this since my gazebo simulation code now relies on this library.
  
Oh, are you saying your gazebo simulation requires an installed "libwave.a" or "libwave.so" and a change to individual object files would disrupt it? I don't object to that. I'm just nitpicking about the internal organization of the project, which can be refactored without affecting the gazebo code. So, this is ready to merge.

chutsu: Indeed, for now I would prefer a single `libwave.a` or `.so`, when the time comes we can revisit this.

PR #64, opened 2017-05-12
Use Doxygen for documentation

### LINKED ISSUE #47: Automate code listings in documentation
Our documentation (see [api_template](https://github.com/wavelab/libwave/blob/master/docs/api/api_template.md)) includes code listings (class definitions, function prototypes, etc). These blocks of code should be automatically generated from the header files, otherwise maintaining the docs is a lot of work (and they will quickly become out of date).

The solution should 
~- be compatible with current ditto system?~
- support Latex, ideally with Mathjax (not image) output

A common solution is [Doxygen](http://www.stack.nl/~dimitri/doxygen/), but will have to investigate whether it works as desired.

(An alternative is to _not_ include code listings in the documentation, and just write "here's a quick example of how to use this class, see header.hpp for the full API".)

leokoppel:
To generate and view the documentation, run:
```bash
sudo apt install doxygen
cd libwave/build;   # (change to whatever your libwave source directory is)
cmake ..
make doc
xdg-open docs/html/index.html  # (open in a web browser)
```

So far the most annoying thing I have noticed with Doxygen is you denote Latex formulas using `@f$` or `\f$`, instead of the usual `$`.

Note the above shows Doxygen's regular html output; @chutsu was working on another way of displaying the output.

Seeking feedback on
- How documentation is organized in the browser
- How documentation is organized in the source code

Resolves #47


chutsu:
I guess we're using javadocs style doxygen? :D
I like it.

leokoppel:
Yes.. I tried https://github.com/sourcey/moxygen before but it didn't seem to support all features.

chutsu:
What specific things would you like support for? My script is also pretty limited.

leokoppel:
@chutsu Here are some things I like in the Doxygen html output:

- Latex formulas printed with MathJax
- API documentation sorted by module
- "Related pages" non-API documentation (for our "fundamentals" pages on theory)
- Function listings with brief descriptions, which link to detailed descriptions
- Links to types and functions mentioned comments
- Links to types in parameter lists
- Links to definitions in source browser
- Inclusion of example code

chutsu:
I don't think I can keep up with parsing doxygen via python in a short amount of time. So, I'll approve of the changes so far assuming we are going towards the doxygen route, at somepoint we will link the doxygen api docs in ditto / revisit creating a script to reproduce the output of doxygen at a later date.

PR #75, opened 2017-05-18
Fix 71 refactor tests

### LINKED ISSUE #71: Refactor test build configuration
Current behaviour:
- Every module has a copy `test_runner.cpp` contaning the `main()` function for a test
- A custom `--silence-stdcout` option is used to silence`std::cout`
- On CI, which tests run is specified in `ci_run.bash` (decoupled from build system)

Suggested changes:
- Configure testing target in CMake
- Use `libgtest-main` which provides a `main` function.
- To silence cout redirect to `/dev/null`

Related issues:
#45, #66

leokoppel:
Resolves #71

- Simplify adding gtests in the CMakeLists: use `wave_add_test` function
- Remove redundant test_runner.cpp files
- Automatically run all defined tests on travis, instead of listing them in a separate file

The behaviour of tests on travis is slightly changed. Instead of printing gtest output (green `RUN...OK`, etc) for each test but silencing other output, it now prints all output only from failed tests. Do you want to see something different?


chutsu:
This point is kind of important to me personally:

>  it now prints all output only from failed tests

I don't know about you but when I'm writing tests I tend to print to screen to verify numbers both in code and visually.

cmake/WaveUtils.cmake
```diff
@@ -0,0 +1,23 @@
+function(wave_add_test NAME)
```
chutsu: I like this, one nit-pick thing though, I personally would like to have a CMake style rule that if its CMAKE commands it should be all caps? Else we now have a mix between lower and ALL CAPS CMake command calls?

leokoppel:
> I don't know about you but when I'm writing tests I tend to print to screen to verify numbers both in code and visually.

@chutsu Sorry, what numbers do you mean? 

There are several types of things that print when you run our google tests

1. Name of each test and result, e.g.
```
[ RUN      ] LandmarkContainer.capacity
[       OK ] LandmarkContainer.capacity (0 ms)
```

2. Failure messages, e.g.
```
.../test.cpp:273: Failure
      Expected: 2ul
      Which is: 2
To be equal to: m2.size()
      Which is: 6
```
3. Stuff your program prints to `cout`

The previous behaviour, when running on travis, is: print 1. and 2. but never 3.

The proposed behaviour is: print 1. 2. and 3. only if the test executable returns an error. If the test executable returns success, print nothing except
```
Start 1: wave_controls_tests
1/6 Test #1: wave_controls_tests ..............   Passed    0.00 sec
```

Which would you like to see? It's easy to change, just would like to clarify.

Also note you can still run an individual test executable as before and see all output 1, 2, and 3. The above only covers the command to run all tests on Travis.

cmake/WaveUtils.cmake
leokoppel: Yes good point, I'll fix it.

chutsu:
Ah, thanks for clarifying, the key point I wanted to make is I want to see the `std::cout` when I am developing on my local machine (3. in your list), similar to how in the gtest mod where I have "--supress-stdcout" when in travis, or let gtest `std::cout` when I am not in travis.

chutsu:
btw did you make a `make tests` target?

chutsu:
nvm found it, `ctest --output-on-failure`

PR #80, opened 2017-05-22
Add range operations to measurement containers

leokoppel:
Continue #41

Add a number of methods working with iterators to MeasurementContainer, to
complete the interface and help with next work (dealing with ranges of landmarks).

New methods:
* getTimeWindow(): get all measurements between two times
* getAllFromSensor(): get all measurements from given sensor id
* constructor taking iterators: make new container with contents of a range
* erase() taking iterators: delete measurements in a range
* insert() taking iterators: insert contents of a range

These can be combined to, for example, copy a range of measurements to another
MeasurementConatiner or to a std::vector.

To help implement the above:
* change underlying sort order
* refactor get() interpolation implementation

Previously, the container was sorted an odd way (sensor id first, then time)
which made this method awkward.  Refactor to sort by time first, then sensor
id, and implement the "complex search" method described at
http://www.boost.org/doc/libs/1_63_0/libs/multi_index/doc/examples.html#example6
for get().

### DIFF
--- a/wave_containers/include/wave/containers/measurement_container.hpp
+++ b/wave_containers/include/wave/containers/measurement_container.hpp
@@ -61,8 +61,8 @@ struct measurement_container {
     struct sensor_key : member<T, decltype(T::sensor_id), &T::sensor_id> {};
 
     // Define a composite key which combines the above two keys
-    // This lets us search elements sorted by sensor id first, then time
-    struct sensor_and_time_key : composite_key<T, sensor_key, time_key> {};
+    // This lets us search elements sorted by time first, then  sensor id
+    struct sensor_and_time_key : composite_key<T, time_key, sensor_key> {};
 
     // These types are used as tags to retrieve each index after the
     // multi_index_container is generated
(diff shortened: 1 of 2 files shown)


wave_containers/include/wave/containers/measurement_container.hpp
```diff
@@ -61,8 +61,8 @@ struct measurement_container {
     struct sensor_key : member<T, decltype(T::sensor_id), &T::sensor_id> {};
 
     // Define a composite key which combines the above two keys
-    // This lets us search elements sorted by sensor id first, then time
-    struct sensor_and_time_key : composite_key<T, sensor_key, time_key> {};
+    // This lets us search elements sorted by time first, then  sensor id
+    struct sensor_and_time_key : composite_key<T, time_key, sensor_key> {};
```
chutsu: Out of curiosity, how likely will it be for 2 different sensors to have exactly the same time key?

leokoppel: How likely will it be for two measurements to have the same time_point but different sensor_id?

Seems pretty likely especially if the user does something like this
```cpp
time_now = getSystemTime();
m.emplace(time_now, sensor1, pollSensor1());
m.emplace(time_now, sensor2, pollSensor2());
```

chutsu: I thought the time has a micro-second resolution?

leokoppel: Probably. But see the example above. It is very plausible that multiple measurements are inserted with the same time; at least, the library should expect it.

Note that is irrelevant to the choice of ordering the index by time first, or by sensor first.

chutsu:
btw you might want to fix the compiler warnings before you merge

screenshot_2017-05-23_23-08-40

PR #83, opened 2017-05-22
Fix 81 implementation files

### LINKED ISSUE #81: Move function template definitions to implementation file
Currently for non-templated code we separate declarations and definitions into separate .hpp and .cpp files, but templated code is in the same file. We [can't put template definitions in .cpp files](https://stackoverflow.com/questions/495021/why-can-templates-only-be-implemented-in-the-header-file)  (unless we only instantiate certain types).

Current approaches:
- Separate declaration and definition, but keep in same file: e.g. [measurement_container.hpp](https://github.com/wavelab/libwave/blob/master/wave_containers/include/wave/containers/measurement_container.hpp)
- Definition inside class: e.g. [factor.hpp PR](https://github.com/wavelab/libwave/pull/57/files#diff-f25a91cb02e7e6b3ddf46bd7a2dd12f8)

Should declarations and definitions be separated?

[Per](https://github.com/wavelab/libwave/pull/57#discussion_r116381251) @arunxdas:
> I think we should. Declarations in header, function body in some kind of impl or cpp file, unless the function body is trivial like a set/get, etc.

I propose the following approach:

In `include/wave/foo/foo.hpp`:
```cpp
template <typename T>
class Foo {
    Foo();
}

// At bottom of file
#include "impl/foo.hpp"
```

In `include/wave/foo/impl/foo.hpp`:
```cpp
template <typename T>
Foo<T>::Foo() {
    // ...
}
```

This approach is used by [kalibr](https://github.com/ethz-asl/kalibr), for example. In terms of code it is the same as [measurement_container.hpp](https://github.com/wavelab/libwave/blob/master/wave_containers/include/wave/containers/measurement_container.hpp), just with a separate file.

We could use a different file extension but I suggest keeping `.hpp`  just because editors recognize it by default. The `impl/` directory makes clear it is an implementation file.

leokoppel:
Resolves #81

For wave_containers:

- Move definitions of function templates into separate "impl" files.
- Move tests for consistency.

Organizational change only; no change in compiled result.


chutsu:
@navganti @Jebediah @arunxdas, thoughts? I personally think it makes it more complicated by putting template declations into a sub `impl` folder, but feel this is a team decision.

leokoppel:
We can discuss possible solutions on the issue page: #81

Jebediah:
I like how this follows the hpp/cpp pattern we have for our other stuff

chutsu:
@navganti @arunxdas ?

Issue #53, opened 2017-04-20
Add exception handling for Rotation class


arunxdas:
Rotation class should have exception handling, specifically to check that input variables are finite, and that the axis angle constructor takes in a normalized rotation axis.

PR #84, opened 2017-05-23
Feature 65 FAST Detector OpenCV Wrapper

navganti:
Pull Request for the start of the vision front-end implementation in libwave. In particular, this is the start for any image feature detectors.

This is a wrapper of OpenCV's FastDetector. The hope is that the base class is modular enough for other feature detectors (GFTT, Shi-Tomasi, etc.) can be implemented in a similar manner. This code will eventually interface with the rest of the front end, feature descriptor modules, and modules to determine feature correspondence and outliers.

Code has been tested + run with cpplint. Docs still to come, as is Doxygen style commenting (once issue #47 is finalized).


leokoppel:
Nice, will review.

I recommend you check out the [rebase](https://git-scm.com/book/en/v2/Git-Branching-Rebasing) command to clean up commit history before PRs.

You can put your comments in Doxygen format even before #47 is finalized.

navganti:
@leokoppel ah, I thought you were still looking to finalize some stuff. I see your comments on #53, will follow that guide.

wave_vision/src/fast_detector.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+FASTDetector::~FASTDetector() {}
+
+void FASTDetector::checkConfiguration(int new_threshold, int new_type) {
+    // Create InvalidConfigurationException object
+    InvalidConfigurationException exception;
+
+    // Check parameters. If invalid, throw an exception.
+    if (new_threshold < 0) {
+        std::cout << "Invalid threshold for FASTDetector!" << std::endl;
```
leokoppel: Don't print to cout.

Usually exception classes take this text as an argument.

For now, don't worry about having your own exception class. Just do something like
```cpp
throw std::invalid_argument("Invalid threshold for FASTDetector");
```

wave_vision/src/fast_detector.cpp
```diff
@@ -0,0 +1,87 @@
+// Libwave Headers
+#include "wave/vision/fast_detector.hpp"
+
+namespace wave {
+
+FASTDetector::FASTDetector(const std::string &config_path) {
+    // Extract parameters from .yaml file.
+    ConfigParser parser;
```
leokoppel: I am skeptical that a "Feature Detector" class should contain a file parser. They are separate concerns.

Certainly a configuration file should not be the only way to initialize a feature detector.

wave_vision/tests/fast_tests.cpp
```diff
@@ -0,0 +1,115 @@
...
+ protected:
+    FASTTest() {}
+    virtual ~FASTTest() {
+        if (this->detector) {
+            delete this->detector;
+        }
+    }
+
+    void initDetector() {
+        this->detector = new FASTDetector(TEST_CONFIG);
```
navganti: This was just for the sake of not having to load the figure over and over again. I ended up only using it for one test, but was trying to follow [this](https://github.com/google/googletest/blob/master/googletest/docs/Primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests). 

Think I may have just done it wrong...haha.

wave_vision/tests/fast_tests.cpp
```diff
@@ -0,0 +1,115 @@
+// Libwave Headers
+#include "wave/wave_test.hpp"
+#include "wave/vision/fast_detector.hpp"
+
+namespace wave {
+
+#define TEST_CONFIG "tests/config/fast.yaml"
+#define TEST_IMAGE "tests/data/lenna.png"
```
chutsu: No, keep Lenna! Its the de-facto example / test image everyone uses, and lets keep that tradition ;P (I'm pretty sure people don't know the significance of Lenna other than us two).

wave_vision/tests/fast_tests.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+    cv::Mat image;
+    FASTDetector *detector;
+    std::vector<cv::KeyPoint> keypoints;
+};
+
+// Checks that correct configuration can be loaded
+TEST(FASTTests, GoodInitialization) {
+    FASTDetector detector(TEST_CONFIG);
+    SUCCEED();
```
navganti: This was more to just confirm that the proper configuration can be loaded. Although...just realized this is redundant due to the "GoodManualConfiguration" test, which loads a proper configuration and then also checks it. Will remove this function.

leokoppel: OK, but what is the test for "proper configuration can be loaded"?

If testing that no exceptions are thrown, use `EXPECT_NO_THROW`.

wave_vision/include/wave/vision/fast_detector.hpp
```diff
@@ -0,0 +1,46 @@
...
+ public:
+    // Public Functions
+    // -------------------------------------------------------------------------
+
+    FASTDetector(const std::string& config_path);
+    ~FASTDetector();
+    void configure(int  new_threshold,
+                   bool new_nonmax_suppression,
+                   int  new_type);
+    void getConfiguration(int&  current_threshold,
```
navganti: FAST only uses these configuration parameters...this would be detector specific.

However, based on your comment below, changing this to an enum could allow me to move this to the FeatureDetector base class, and extract enum parameters? The actual implementation would still have nuances from detector to detector.

wave_vision/src/fast_detector.cpp
navganti: No, it isn't. The detector could also be created by the user manually setting the configuration parameters. I decided to use a config file in order to maintain similarity with @Jebediah's scan matching interface.

wave_vision/include/wave/vision/fast_detector.hpp
leokoppel: Comments below are unrelated. I mean, it is clumsy to have a single "getConfiguration" function with a parameter for each value. You could use individual getters / setters, or pass a struct.

wave_vision/src/fast_detector.cpp
leokoppel: Sorry if I missed it; I only see one constructor and it takes a filename.

wave_vision/src/fast_detector.cpp
```diff
@@ -0,0 +1,87 @@
...
+    }
+}
+
+void FASTDetector::configure(int new_threshold,
+                                     bool new_nonmax_suppression,
+                                     int new_type) {
+    // Confirm configuration parameters are valid
+    checkConfiguration(new_threshold, new_type);
+
+    // Store configuration values in member variables
```
leokoppel: > // Store configuration values in member variables
 > // Set configuration values in detector.

Is this not redundant?

wave_vision/src/fast_detector.cpp
navganti: No no, you're right. I didn't include that implementation. I will add two more constructors; a default one and one which takes in the configuration arguments directly.

wave_vision/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+ADD_DEFINITIONS(${EIGEN_DEFINITIONS}
+                ${OpenCV_DEFINITIONS})
+
+# INCLUDES
+INCLUDE_DIRECTORIES(
+    ${EIGEN_INCLUDE_DIRS}
+    ${OpenCV_INCLUDE_DIRS})
+
+# LIBRARY DIRS
+LINK_DIRECTORIES(${EIGEN_LIBRARY_DIRS})
```
navganti: Fixed.

wave_vision/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+# PACKAGES
+FIND_PACKAGE(OpenCV REQUIRED)
+
+# DEFINITIONS
+ADD_DEFINITIONS(${EIGEN_DEFINITIONS}
+                ${OpenCV_DEFINITIONS})
+
+# INCLUDES
+INCLUDE_DIRECTORIES(
+    ${EIGEN_INCLUDE_DIRS}
```
navganti: fixed.

wave_vision/src/fast_detector.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+    // Add parameters to parser, to be loaded. If path cannot be found, throw an
+    // exception.
+    parser.addParam("fast.threshold", &param_threshold);
+    parser.addParam("fast.nonmax_suppression", &param_nonmax_suppression);
+    parser.addParam("fast.type", &param_type);
+
+    if (parser.load(config_path) != 0) {
+        ConfigurationLoadingException exception;
+        throw exception;
```
navganti: done.

wave_vision/src/fast_detector.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+std::vector<cv::KeyPoint>& FASTDetector::detectFeatures(const cv::Mat& image) {
+    // Load image
+    this->loadImage(image);
+
+    // Detect features in image, save values into keypoints.
+    this->fast_detector->detect(this->image, this->keypoints);
+
+    return this->keypoints;
```
navganti: I made them as a member just in case...for potential future features. For example, in @chutsu's yarl library, he has a function to convert from cv::KeyPoint to cv::Point2f (because RANSAC apparently requires that format). 

I figured I'd keep it in case future operations with the keypoints need to be implemented. If there's no use for it at that point it can be removed.

wave_vision/tests/fast_tests.cpp
navganti: fixed

wave_vision/tests/fast_tests.cpp
navganti: fixed.

wave_vision/include/wave/vision/fast_detector.hpp
navganti: changed, please take a look at this new implementation. I removed the configuration member variables, as it seemed no longer necessary.

wave_vision/src/fast_detector.cpp
navganti: Constructors added.

wave_vision/src/fast_detector.cpp
navganti: Configuration parameters are no longer a member variable, this concern should be addressed now.

wave_vision/tests/fast_tests.cpp
```diff
@@ -0,0 +1,115 @@
+// Libwave Headers
+#include "wave/wave_test.hpp"
+#include "wave/vision/fast_detector.hpp"
+
+namespace wave {
+
+#define TEST_CONFIG "tests/config/fast.yaml"
```
navganti: done

navganti:
Hi @leokoppel, I think the last thing I need to do is Doxygen on the main fast_detector.cpp, but I've made most of the changes you've requested.

wave_vision/include/wave/vision/feature_detector.hpp
```diff
@@ -0,0 +1,81 @@
...
+
+/** C++ Headers */
+#include <exception>
+#include <vector>
+
+/** Libwave Headers */
+#include "wave/utils/utils.hpp"
+
+/** Third Party Headers */
+#include "opencv2/opencv.hpp"
```
leokoppel: Since you're being a stickler about cpplint-enforced #includes, isn't this the wrong order?

[Names and Order of Includes](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes):

> 1.  dir2/foo2.h.
> 2. C system files.
> 3. C++ system files.
> 4. Other libraries' .h files.
> 5. Your project's .h files.

Use <> for opencv headers

wave_vision/src/fast_detector.cpp
leokoppel: It is not part of the public interface, so there's no point in having it "just in case". (I don't see why they would be need to stored as a member anyway, but if so you can add it later.)

wave_vision/tests/fast_tests.cpp
```diff
@@ -0,0 +1,117 @@
...
+    ASSERT_EQ(output_config.type, input_config.type);
+}
+
+// Confirms that keypoints can be determined, and image with keypoints can be
+// displayed.
+TEST_F(FASTTest, DetectImage) {
+    this->keypoints = detector.detectFeatures(this->image);
+    ASSERT_NE(this->keypoints.size(), 0);
+
+    // VISUAL TEST TO CONFIRM IMAGE AND KEYPOINTS ARE DISPLAYED PROPERLY
```
leokoppel: Avoid submitting commented code.

A quick fix is to [disable the test](https://github.com/google/googletest/blob/48ee8e98abc950abd8541e15550b18f8f6cfb3a9/googletest/docs/AdvancedGuide.md#temporarily-disabling-tests) by naming it `DISABLED_DetectImage`. Then the test will be compiled, just not run.

wave_vision/src/fast_detector.cpp
```diff
@@ -0,0 +1,98 @@
...
+    // Confirm configuration parameters are valid
+    checkConfiguration(new_config);
+
+    // Set configuration values in detector.
+    this->fast_detector->setThreshold(new_config.threshold);
+    this->fast_detector->setNonmaxSuppression(new_config.nonmax_suppression);
+    this->fast_detector->setType(new_config.type);
+}
+
+FASTParams FASTDetector::getConfiguration() {
```
leokoppel: Nice, this is much better.

wave_vision/include/wave/vision/fast_detector.hpp
```diff
@@ -0,0 +1,106 @@
...
+ * [opencv_feature_detectors]: http://docs.opencv.org/trunk/df/d74/classcv_1_1FastFeatureDetector.html
+ */
+class FASTDetector : public FeatureDetector {
+ public:
+    /** Default Constructor */
+    FASTDetector();
+
+    /** Constructs a FastFeatureDetector using parameters specified by the user
+     *
+     *  @param config, a FASTParams struct containing the desired threshold,
```
leokoppel: Nice. By the way, it is not strictly necessary to list all the @params or @returns if they are obvious.
It is never necessary to mention the type of the parameter, because it will appear in Doxygen with a hyperlink to the type's documentation.

wave_vision/include/wave/vision/fast_detector.hpp
```diff
@@ -0,0 +1,106 @@
...
+/** @addtogroup vision
+ *  @{ */
+
+/**
+ *  FASTParams struct contains the configuration parameters for the
+ *  FastFeatureDetector.
+ *
+ */
+struct FASTParams {
+    int threshold;
```
leokoppel: Move comments from elsewhere about each member (e.g. possible values) here

wave_vision/include/wave/vision/fast_detector.hpp
```diff
@@ -94,9 +98,9 @@ class FASTDetector : public FeatureDetector {
     *  @param check_config, the FASTParams struct containing the desired
     *  configuration values.
     */
-    void checkConfiguration(FASTParams& check_config);
+    void checkConfiguration(const FASTParams& check_config);
 };
 
-} // end of namespace wave
+}  // end of namespace wave
```
leokoppel: By the way is your cpplint ok with this? It wants me to use "// namespace wave"

chutsu: ah right, I'm not even using cpplint yet. Do you have a config file for that? by default cpplint adheres to Google's style guide no?

leokoppel: Yes it's only purpose is enforcing google style guide. We're not enforcing it in this project right? I just happened to have it on in my IDE.

wave_vision/include/wave/vision/feature_detector.hpp
navganti: ha I guessed, and cpplint didn't complain about it. Fixed now

wave_vision/include/wave/vision/fast_detector.hpp
chutsu: ah, I still think we can tune cpplint to our desire, for example to ignore the namespace warning you can create a `CPPLINT.cfg` config file with the following:

    filter=-readability/namespace

to disable the `// namespace wave` warning.

wave_vision/include/wave/vision/fast_detector.hpp
navganti: done

wave_vision/tests/fast_tests.cpp
navganti: removed.

wave_vision/include/wave/vision/fast_detector.hpp
navganti: It didn't complain about it to me

navganti: How are you running cpplint @leokoppel?

PR #88, opened 2017-05-23
Feature 79 add covar

Jebediah:
This PR adds two options to calculate an estimated information matrix for scan registration using ICP:
- Lu and Milios
- Censi

At a high level, LUM is a bit faster but does not account for sensor noise. Censi takes a bit longer but includes a sensor model. Here, the model implemented is to assume each point is independent and that there is Gaussian noise on the measured angles and range. The magnitude of that noise is arbitrary (until we measure it) but is loosely based on our Lidar specs.

Note that an important caveat is that these methods do not take into account incorrect correspondences (too slow for realtime) so they cannot detect incorrect matches by themselves.


wave_matching/include/wave/matching/icp.hpp
```diff
@@ -15,11 +15,21 @@ class ICPMatcher : public Matcher<PCLPointCloud> {
     void setRef(const PCLPointCloud &ref);
     void setTarget(const PCLPointCloud &target);
     bool match();
+    void estimate_info();
```
chutsu: camelCase

wave_matching/src/icp.cpp
```diff
@@ -17,11 +17,15 @@ ICPMatcher::ICPMatcher(float res, const std::string &config_path) {
...
-    int max_iter;
+    int max_iter, estimate_method_int;
     parser.addParam("icp.max_corr", &max_corr);
     parser.addParam("icp.max_iter", &max_iter);
     parser.addParam("icp.t_eps", &t_eps);
     parser.addParam("icp.fit_eps", &fit_eps);
+    parser.addParam("icp.lidar_ang_covar", &(this->lidar_ang_covar));
+    parser.addParam("icp.lidar_lin_covar", &(this->lidar_lin_covar));
+    parser.addParam("icp.covar_estimator", &estimate_method_int);
+    this->estimate_method = static_cast<ICPMatcher::covar_method> (estimate_method_int);
```
chutsu: Put this after `parser.load` because the `estimate_method_int` doesn't contain the parameter value until the parameter is loaded?

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
     return false;
 }
 
+void ICPMatcher::estimate_info() {
```
chutsu: camelCase

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+    switch (this->estimate_method) {
+        case ICPMatcher::covar_method::LUM: this->estimate_lum();
+        case ICPMatcher::covar_method::CENSI: this->estimate_censi();
+        default: return;
+    }
+}
+
+// Taken from the Lu and Milios matcher in PCL
+void ICPMatcher::estimate_lum() {
```
chutsu: camelCase

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+    if (this->icp.hasConverged()) {
+        auto list = this->icp.correspondences_.get();
+        Mat6 MM = Mat6::Zero ();
+        Vec6 MZ = Vec6::Zero ();
+        std::vector <Eigen::Vector3f> corrs_aver;
+        std::vector <Eigen::Vector3f> corrs_diff;
+        int numCorr = 0;
```
chutsu: So dense, new lines!!!

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+}
+
+// Taken from the Lu and Milios matcher in PCL
+void ICPMatcher::estimate_lum() {
+    if (this->icp.hasConverged()) {
+        auto list = this->icp.correspondences_.get();
+        Mat6 MM = Mat6::Zero ();
+        Vec6 MZ = Vec6::Zero ();
+        std::vector <Eigen::Vector3f> corrs_aver;
+        std::vector <Eigen::Vector3f> corrs_diff;
```
chutsu: Why floats and not doubles?

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+                corrs_diff.push_back(Eigen::Vector3f(
+                        this->ref->points[it->index_query].x - this->target->points[it->index_match].x,
+                        this->ref->points[it->index_query].y - this->target->points[it->index_match].y,
+                        this->ref->points[it->index_query].z - this->target->points[it->index_match].z));
+                numCorr++;
+            }
+        }
+
+        for (int ci = 0; ci != numCorr; ++ci)  // ci = correspondence iterator
+        {
```
chutsu: clang-format?

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+            MM (4, 4) += corrs_aver[ci] (0) * corrs_aver[ci] (0) + corrs_aver[ci] (1) * corrs_aver[ci] (1);
+            MM (5, 5) += corrs_aver[ci] (0) * corrs_aver[ci] (0) + corrs_aver[ci] (2) * corrs_aver[ci] (2);
+
+            // Fast computation of M'Z
+            MZ (0) += corrs_diff[ci] (0);
+            MZ (1) += corrs_diff[ci] (1);
+            MZ (2) += corrs_diff[ci] (2);
+            MZ (3) += corrs_aver[ci] (1) * corrs_diff[ci] (2) - corrs_aver[ci] (2) * corrs_diff[ci] (1);
+            MZ (4) += corrs_aver[ci] (0) * corrs_diff[ci] (1) - corrs_aver[ci] (1) * corrs_diff[ci] (0);
+            MZ (5) += corrs_aver[ci] (2) * corrs_diff[ci] (0) - corrs_aver[ci] (0) * corrs_diff[ci] (2);
```
chutsu: clang-format?

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+                                      + pow (corrs_diff[ci] (1) - (D (1) + corrs_aver[ci] (0) * D (4) - corrs_aver[ci] (2) * D (3)), 2.0f)
+                                      + pow (corrs_diff[ci] (2) - (D (2) + corrs_aver[ci] (1) * D (3) - corrs_aver[ci] (0) * D (5)), 2.0f));
+        }
+
+        // When reaching the limitations of computation due to linearization
+        if (ss < 0.0000000000001 || !pcl_isfinite (ss))
+        {
+            this->information = Mat6::Identity();
+            return;
+        }
```
chutsu: clang-format?

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+// The core idea behind this is that the covariance of the cost f'n J wrt
+// optimization variable x is
+// cov(x) ~= (d2J/dx2)^-1*(d2J/dzdx)*cov(z)*(d2J/dzdx)'*(d2J/dx2)^-1
+
+// Idea was taken from
+// https://censi.science/pub/research/2007-icra-icpcov.pdf
+
+// This is an implementation for euler angles, what is below is a cleaned up
+// version of
+// http://ieeexplore.ieee.org/stamp/stamp.jsp?arnumber=7153246
```
chutsu: Not quite a doxygen style comment, we following something very similar to http://mesos.apache.org/documentation/latest/doxygen-style-guide/

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+//                [               0,                   0,                  0, sin(S5)*sin(S6),  S4*cos(S5)*sin(S6), S4*cos(S6)*sin(S5)]
+//                [               0,                   0,                  0,         cos(S6),                   0,        -S4*sin(S6)]
+
+                // d2J_dx2
+
+                d2J_dX2(0, 0) += 2;
+                d2J_dX2(1, 1) += 2;
+                d2J_dX2(2, 2) += 2;
+
+                d2J_dX2(0, 3) += 2 * Z2 * (sr * sy + cr * cy * sp) +
```
chutsu: consider switching off clang-format by `// clang-format off` for readability, and switch it back on after this giant block of code with `// clang-format on`

wave_matching/src/icp.cpp
```diff
@@ -75,4 +79,367 @@ bool ICPMatcher::match() {
...
+                                 2 * Z1 * cp * sy;
+                d2J_dZdX(4, 5) = 2 * Z2 * (cr * sy - cy * sr * sp) -
+                                 2 * Z3 * (sr * sy + cr * cy * sp) -
+                                 2 * Z1 * cp * cy;
+                // d2J_dZdX(5,5) = 0;
+
+                middle.noalias() += d2J_dZdX * cov_Z * (d2J_dZdX.transpose());
+            }
+        }
+        Eigen::MatrixXd inverse =
```
chutsu: MatX

wave_matching/src/icp.cpp
Jebediah: This bit was taken from PCL and they used floats. Don't think it really matters one way or the other, the extra precision of double isn't going to change anything.

wave_matching/src/icp.cpp
Jebediah: The whole matcher actually needs to be updated with doxygen docs. I'll do that in a seperate PR.

wave_matching/tests/ndt_tests.cpp
```diff
@@ -5,8 +5,8 @@
 
 namespace wave {
 
-#define TEST_SCAN "tests/data/testscan.pcd"
-#define TEST_CONFIG "tests/config/ndt.yaml"
+#define TEST_SCAN "data/testscan.pcd"
+#define TEST_CONFIG "config/ndt.yaml"
```
chutsu: probably want to do a `static const std::string TEST_CONFIG = "config/ndt.yaml"`

The reason is because the `#define TEST_CONFIG` is defined in other test files, this is working without modification because this line overwrites the previous `TEST_CONFIG` macro. Ideally we want to: 

1. avoid the use of macros (leo will bring this up, as he already has with #84 [example](https://github.com/wavelab/libwave/blob/master/wave_vision/tests/fast_tests.cpp)). 
2. avoid name conflicts

One solution is to use `static` to limit the scope of the variable to only this file, and `const` because we're going to set it once and never change again. Hence the

    static const std::string TEST_CONFIG = "config/ndt.yaml"

wave_matching/src/icp.cpp
chutsu: ok

Jebediah: #112

wave_matching/tests/ndt_tests.cpp
leokoppel: Use `const auto TEST_CONFIG = "config/ndt.yaml"`

PR #89, opened 2017-05-23
Feature 59 trajectory error

Jebediah:
This is to add a small utility class to calculate error over a SLAM trajectory given ground truth. Uses our new measurement containers to interpolate between ground truth points.


wave_benchmark/include/wave/benchmark/pose_measurement.hpp
```diff
@@ -0,0 +1,49 @@
...
+    m1inverse.invert();
+    auto relative = m2.value.rotation * m1inverse;
+    wave::Vec3 wvec = w2 * relative.logMap();
+    wave::Rotation interpolated;
+    interpolated.setFromExpMap(wvec);
+    auto trans = (1 - w2) * m1.value.translation + w2 * m2.value.translation;
+    Pose retval;
+    retval.rotation = interpolated * m1.value.rotation;
+    retval.translation = trans;
+    return retval;
```
leokoppel: Looks good

wave_benchmark/include/wave/benchmark/trajectory_compare.hpp
```diff
@@ -0,0 +1,33 @@
...
+    void push_truth(Pose pose, const TimeType time);
+    void push_measurement(Pose pose, const TimeType time);
+    void calculate_error();
+    void output_csv(std::string path, std::string filename);
+    MeasurementContainer<PoseMeasurement> ground_truth;
+    MeasurementContainer<PoseMeasurement> measurements;
+    MeasurementContainer<PoseMeasurement> error;
+
+ private:
+    const int ground_truth_key = 0;
```
leokoppel: What's the point of these ints?

wave_benchmark/include/wave/benchmark/trajectory_compare.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <iostream>
+#include <fstream>
+
+#include "pose_measurement.hpp"
+#include "wave/containers/measurement_container.hpp"
+#include "wave/geometry/rotation.hpp"
+
+namespace wave {
+
+class TrajectoryCompare {
```
leokoppel: Add docs (in doxygen format)

wave_benchmark/src/trajectory_compare.cpp
```diff
@@ -0,0 +1,44 @@
...
+        truth.rotation.invert();
+        error.rotation = truth.rotation * iter->value.rotation;
+        error.translation = iter->value.translation - truth.translation;
+        this->error.insert(
+          PoseMeasurement(iter->time_point, this->error_key, error));
+    }
+}
+
+void TrajectoryCompare::output_csv(std::string path, std::string filename) {
+    file.open(path + filename);
```
leokoppel: Missing `this->`; should not be a member anyway

leokoppel: Why take two parameters only to concatenate them. Take just one parameter. Take `const std::string &`

wave_benchmark/src/trajectory_compare.cpp
```diff
@@ -0,0 +1,44 @@
...
+namespace wave {
+
+void TrajectoryCompare::reset() {
+    this->ground_truth.clear();
+    this->measurements.clear();
+    this->error.clear();
+}
+
+void TrajectoryCompare::push_measurement(Pose pose, const TimeType time) {
+    measurements.insert(PoseMeasurement(time, this->measurement_key, pose));
```
leokoppel: Use `emplace(...)`

wave_benchmark/src/trajectory_compare.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void TrajectoryCompare::push_truth(Pose pose, const TimeType time) {
+    ground_truth.insert(PoseMeasurement(time, this->ground_truth_key, pose));
+}
+
+void TrajectoryCompare::calculate_error() {
```
leokoppel: This function looks good. In fact, it should be more prominent. It is the meat of the TrajectoryCompare class. The whole interface should be designed around it. 

 When someone wants to calculate error, what do they typically have? Two sequences of measurements.

They already have two sequences of measurements. Why should they have to learn how to feed the measurements-by-one into a `TrajectoryCompare` class, just so they can get at the juicy meat of `calculateError()`?

---

Here is the problem and solution in my view. You have two sequences of poses. You want to calculate the error. You want to print output to a csv file. What do you _really_ need?

```cpp
error_poses calculateError(estimate_poses, truth_poses);
void outputToCsv(poses, file path);
```

Do you really need the rest?

wave_benchmark/include/wave/benchmark/pose_measurement.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+#include <chrono>
+#include "wave/geometry/rotation.hpp"
+#include "wave/utils/math.hpp"
+
+namespace wave {
+
+using TimeType = std::chrono::steady_clock::time_point;
+
+struct Pose {
```
Jebediah: Maybe, need to discuss. I created this version of Pose to suit my needs only.

wave_benchmark/include/wave/benchmark/pose_measurement.hpp
```diff
@@ -0,0 +1,49 @@
...
+    wave::Rotation rotation;
+    wave::Vec3 translation;
+};
+
+struct PoseMeasurement {
```
Jebediah: Not really. The ground truth is probably coming from GPS but it doesn't have to. They are just generic poses.

wave_benchmark/include/wave/benchmark/trajectory_compare.hpp
Jebediah: Just to have something consistent for the sensor ids.

wave_benchmark/include/wave/benchmark/pose_measurement.hpp
```diff
@@ -0,0 +1,49 @@
+#ifndef WAVE_BENCHMARK_POSE_MEASUREMENT_HPP
+#define WAVE_BENCHMARK_POSE_MEASUREMENT_HPP
+
+#include <chrono>
+#include "wave/geometry/rotation.hpp"
+#include "wave/utils/math.hpp"
+
+namespace wave {
+
+using TimeType = std::chrono::steady_clock::time_point;
```
chutsu: If we're not careful would this not conflict with the measurement container's TimeType ([here](https://github.com/wavelab/libwave/blob/master/wave_containers/include/wave/containers/measurement_container.hpp#L28))

wave_benchmark/include/wave/benchmark/pose_measurement.hpp
```diff
@@ -0,0 +1,49 @@
...
+};
+
+struct PoseMeasurement {
+    TimeType time_point;
+    int sensor_id;
+    Pose value;
+    PoseMeasurement(const TimeType &t, const int &s, const Pose &v)
+        : time_point{t}, sensor_id{s} {
+        value = v;
+    }
```
chutsu: Have you found a fix for not needing to put `value = v` in the constructor body? Ideally you want to avoid that if possible.

wave_benchmark/include/wave/benchmark/trajectory_compare.hpp
```diff
@@ -0,0 +1,33 @@
...
+ public:
+    void reset();
+    void push_truth(Pose pose, const TimeType time);
+    void push_measurement(Pose pose, const TimeType time);
+    void calculate_error();
+    void output_csv(std::string path, std::string filename);
```
chutsu: camelCases for Class methods?

wave_benchmark/src/trajectory_compare.cpp
Jebediah: That's a good point. I will add that but I think I'm going to keep the one-by-one feeding just because it is convenient if the measurements are coming from a rosbag: The rosbag API isn't well documented and the only way I know to consume them is to iterate through, so 1-by-1 feeding is natural.

wave_benchmark/src/trajectory_compare.cpp
```diff
@@ -0,0 +1,44 @@
...
+        Pose error;
+        truth.rotation.invert();
+        error.rotation = truth.rotation * iter->value.rotation;
+        error.translation = iter->value.translation - truth.translation;
+        this->error.insert(
+          PoseMeasurement(iter->time_point, this->error_key, error));
+    }
+}
+
+void TrajectoryCompare::output_csv(std::string path, std::string filename) {
```
chutsu: Would be very useful while you're filling up the doxygen comment that you mention what the format of the CSV file is, also I don't see a comma while you write the csv file?

wave_benchmark/src/trajectory_compare.cpp
Jebediah: Although what is the datatype being passed in? measurement containers?

wave_benchmark/src/trajectory_compare.cpp
Jebediah: Okay. Only reason I used insert is because I didn't know what kind of argument emplace took.
```
template <typename T>
template <typename... Args>
std::pair<typename MeasurementContainer<T>::iterator, bool>
MeasurementContainer<T>::emplace(Args &&... args) {
// Support Boost.MultiIndex <= 1.54, which does not have emplace()
#if BOOST_VERSION < 105500
    return this->composite().insert(
      MeasurementType{std::forward<Args>(args)...});
#else
    return this->composite().emplace(std::forward<Args>(args)...);
#endif
}
```

wave_benchmark/include/wave/benchmark/trajectory_compare.hpp
leokoppel: If you have a small set of possible values, not representing a quantity, use an enum.

wave_benchmark/src/trajectory_compare.cpp
chutsu: If I'm not (too lazy to double check), think of emplace as a short hand direct access to the objects constructor, the `measurements` already knows its storing `PoseMeasurements` therefore using emplace you don't have to write

    measurements.insert(PoseMeasurement(time, ....

emplace allows you to simplify it to 

        measurements.emplace(time, ....

wave_benchmark/src/trajectory_compare.cpp
leokoppel: Yes it could be a measurement container. It could also be something more flexible like a pair of iterators for each sequence.

Sure, 1-by-1 iteration makes sense but it is simpler for the user (and much simpler for you, the maintainer) to build up a standard container first. It is also much simpler in the plausible event that the user already has a container.

wave_benchmark/include/wave/benchmark/pose_measurement.hpp
```diff
@@ -9,6 +9,7 @@
 #include "wave/geometry/rotation.hpp"
 #include "wave/utils/math.hpp"
 #include "wave/containers/measurement.hpp"
+#include "enum_class.hpp"
```
chutsu: `wave/benchmark/enum_class`

Jebediah:
@leokoppel @chutsu Could you guys update your reviews?

leokoppel:
Hi, sorry about not updating. My one major change request still applies.

wave_benchmark/include/wave/benchmark/pose_measurement.hpp
chutsu: typo?

leokoppel:
Merge this if desired but it will require more changes to be useful in general.

The `ComparisonKey` type is not very useful, and having `wave::PoseMeasurement` harcoded to use it drastically reduces the usefulness of this class. There's a good chance we will want a more general-purpose `wave::PoseMeasurement` in the future.

Jebediah:
Yeah, but I think that our pose measurement type will pop out of the optimiser work, so might as well wait.

leokoppel:
Yep, good idea

Issue #90, opened 2017-05-23
Reconfigure Travis notifcations to slack instead of emailing Chris


chutsu:
At current whenever a branch fails or gets fixed @chutsu gets notified. This issue aims to

- Remove Chris from travis notications
- Redirect travis notifications to slack

leokoppel:
It would be nice to limit this so it doesn't show build failures on topic branches on slack. Ideally it would notify only for "cron" builds, so we have an alert if master is not building due to some changed depedency.

Unfortunately there's no way to limit it to a specific branch through travis.yml: https://github.com/travis-ci/travis-ci/issues/1405

Maybe:
 - disable notifications
 - use a separate slack channel?

chutsu:
Actually I'm for disabling notifications actually

PR #91, opened 2017-05-24
Fix 90 notifications

chutsu:
In relation to #90 

This PR removes travis notifications so that I don't get emails for every branch failure. I don't think we care too much about getting travis notifications since we already have a travis check before a PR can be merged in.


leokoppel:
I am getting emails for my own commits now.

PR #122, opened 2017-05-31
Initial factor and variable implementation

### LINKED ISSUE #44: FactorGraph Container
The goal of this issue is to implement a GTSAM like factor graph container that would turn a factor graph optimization problem and map it to a Ceres solver problem or other external optimizer of our choosing.

leokoppel:
Closes #44 

Based on the draft in #57.

This design aims to solve the problem of
  - implementing a factor graph,
  - with heterogeneous n-ary factors and variables,
  - that must eventually be evaluated in the form of [Ceres' CostFunction](http://ceres-solver.org/nnls_modeling.html#costfunction),
  - while retaining "nice" modern syntax and strong type-checking for the residual and Jacobian calculation functions

This PR does not do too much with the Graph, and mainly implements Factors and Variables. It includes one sample Factor type and two sample Variable types, defined in `wave_optimization/tests/factor_graph/example_instances.hpp`. These classes are demonstrated in `wave_optimization/tests/factor_graph/factor_test.cpp`.

For this PR I encourage you to examine the above two files and give feedback on the form and use of the Factor and Variable classes. Should you look at the implementation details of the base `Factor` template, be warned there is some ugly metaprogramming behind the scenes (its purpose is converting Ceres function calls to "nice" ones).

See my detailed notes on "[SLAM] Ceres Interface" board.


leokoppel:
Oh, OK...

PR #123, opened 2017-05-31
Refactor measurement containers

leokoppel:
Further to #41

Reduce code duplication between the similar MeasurementContainer and LandmarkMeasurementContainer (and any other future variations).

This is done by implementing static polymorphism, using CRTP (the technique used extensively by Eigen). Resources:
- http://eli.thegreenplace.net/2011/05/17/the-curiously-recurring-template-pattern-in-c
- https://stackoverflow.com/questions/4173254/what-is-the-curiously-recurring-template-pattern-crtp
- https://stackoverflow.com/questions/6006614/c-static-polymorphism-crtp-and-using-typedefs-from-derived-classes

#### Pre-Merge Checklist:
- [x] Zero compiler warnings

Testing done: existing unit tests. There is no behaviour change, so unit tests are untouched.


wave_containers/include/wave/containers/impl/landmark_measurement_container.hpp
```diff
@@ -19,7 +19,11 @@ using boost::multi_index::tag;
  * holding measurements of type T. See `wave::internal::measurement_container`.
  */
 template <typename T>
-struct landmark_container {
+struct container_traits<::wave::LandmarkMeasurementContainer<T>> {
+    // Define to give MeasurementContainerBase access to its derived type
+    using MeasurementType = T;
```
leokoppel: Yes. `T` does not persist outside of this definition. `MeasurementType` is defined so it can be used by `MeasurementContainerBase`, from outside of this definition.

navganti: Ah, right...because Base is templated with either the landmark or regular measurement containers.

I was actually wondering about where MeasurementType was defined in the base class; that makes a bit more sense now. Thanks

wave_containers/include/wave/containers/impl/measurement_container_base.hpp
```diff
@@ -0,0 +1,161 @@
...
+#else
+    return this->composite().emplace(std::forward<Args>(args)...);
+#endif
+}
+
+template <typename Derived>
+template <typename... Args>
+typename MeasurementContainerBase<Derived>::size_type
+MeasurementContainerBase<Derived>::erase(Args &&... args) {
+    auto &composite = this->composite();
```
leokoppel: Mainly for readability. I could call `this->composite()` each time, but that's
- longer
- unclear whether the call is the same each time / has side effects

wave_containers/include/wave/containers/measurement_container_base.hpp
```diff
@@ -0,0 +1,195 @@
...
+    // Iterators
+
+    iterator begin() noexcept;
+    iterator end() noexcept;
+    const_iterator begin() const noexcept;
+    const_iterator end() const noexcept;
+    const_iterator cbegin() const noexcept;
+    const_iterator cend() const noexcept;
+
+ protected:
```
leokoppel: Hmm maybe. I sorted them by types and data members but I actually have a typedef down here.. good catch

wave_containers/include/wave/containers/impl/measurement_container_base.hpp
leokoppel: So the next line doesn't go over 80 characters ; )

But also, it makes it clear I just use the result of one call `composite()`; that it doesn't change or have side effects.

leokoppel: To keep the next line under 80 columns ; )

But also, it makes it clear I only need one result of `composite()`; that it doesn't change or have side effects.

wave_containers/include/wave/containers/measurement_container_base.hpp
leokoppel: Removed the typedef at the bottom (and the `protected` at the top)

wave_containers/include/wave/containers/impl/measurement_container_base.hpp
navganti: Okay, sounds good. Was just curious since you were just directly modifying this->composite all throughout the rest of those similar methods.

Issue #119, opened 2017-05-30
Add VO synthetic data set generator


chutsu:
For #98 a VO synthetic data set generator is required. This PR implements a 2 wheel robot going in a circle and output the the observed and ground truth landmarks in the image and world frames, as well as the robot pose.

PR #127, opened 2017-06-01
Add VO synthetic dataset generator

chutsu:
Related to #119 

This PR implements a VO synthetic dataset generator. The generator simulates a two wheel robot going round in a circle and generates csv files that record:

- Robot pose
- 3D world features
- Observed features in camera frame

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


wave_vision/CMakeLists.txt
```diff
@@ -4,40 +4,52 @@ PROJECT(wave_vision)
 # DEPENDENCIES
 FIND_PACKAGE(OpenCV REQUIRED)
 FIND_PACKAGE(Eigen3 REQUIRED)
+FIND_PACKAGE(Boost 1.40 COMPONENTS system filesystem REQUIRED)
```
leokoppel: Hmm, why do you require Boost 1.40?

wave_vision/src/dataset.cpp
```diff
@@ -0,0 +1,255 @@
...
+    Vec4 f_3d, f_3d_edn;
+    Mat3 R;
+    MatX P;
+
+    // pre-check
+    if (this->update(dt) == false) {
+        return 1;
+    }
+
+    // rotation matrix - convert from nwu to edn then to rotation matrix R
```
chutsu: It matters yes, because in text books when we talk about projection matrices with R and t, they assume you're using the camera coordinate system (EDN: x - right, y - down, z - forward), however, I'm using a Two Wheel Robot to traverse the world, a standard coordinate system for simulators like Gazebo is (NWU: x - forward, y - left, z - up). 

So the tldr; Conversions are necessary if we want to use something like Gazebo simulator to extend this, or we can refactor this and make everything in camera coordinate system (EDN) that would reduce the conversions in this code.

chutsu: I should have the conversion code in the `TestDataset::generateDataset()` function though.

wave_vision/CMakeLists.txt
chutsu: Because C++ suprisingly doesn't have a  `rmdir`. I found sources that strongly recommended for `boost::filesystem::remove_all()` function. But I found another implementation that uses standard C++. Will remove this.

In regards to the FindBoost module, do you think it is a good policy to have our own copies of the modules even though they may exist on the system?

wave_vision/src/dataset.cpp
navganti: ah okay. We've talked before about NWU being the standard. I was more curious about why the change to EDN - Gazebo makes sense.

Hmm, up to you. I don't think it makes too much of a difference. As long as we (and the future users) know to keep track of everything it should be okay. Coordinate systems can quickly mess up a lot of things!

wave_vision/CMakeLists.txt
leokoppel: I have no problem with using Boost, the question is why 1.40? It implies our software will work with any version >= 1.40 which I doubt is true (we test 1.54).

I think we should maintain as few scripts as possible.

leokoppel: @chutsu in fact you _should_ use boost filesystem, no point maintaining our own non-cross-platform filesystem code.

chutsu: -____-

leokoppel: Again, my question was solely about this CMakeLists command.

wave_vision/src/dataset.cpp
```diff
@@ -0,0 +1,253 @@
...
+    MatX P;
+    projection_matrix(this->K, R, -R * t, P);
+
+    // check which features in 3d are observable from camera
+    observed.clear();
+    for (int i = 0; i < features.cols(); i++) {
+        // convert feature in NWU to EDN coordinate system
+        Vec4 f_3d, f_3d_edn;
+        f_3d = features.block(0, i, 4, 1);
+        f_3d_edn(0) = -f_3d(1);
```
navganti: @chutsu I see you've added the NWU to EDN function above...why are you doing this manually here again?

chutsu: This is because the NWU to EDN functions assume the vectors are of size 3, in this case the vector are size 4 so I can't reuse the defined functions.

navganti: Okay. This looks good to me.

PR #137, opened 2017-06-07
Base factor graph implementation

### LINKED ISSUE #44: FactorGraph Container

leokoppel:
Closes #44 
Precursors: #57, #122

This design aims to solve the problem of
  - implementing a factor graph,
  - with heterogeneous n-ary factors and variables,
  - that must eventually be evaluated in the form of [Ceres' CostFunction](http://ceres-solver.org/nnls_modeling.html#costfunction),
  - while retaining "nice" modern syntax and strong type-checking for the residual and Jacobian calculation functions

This PR defines base classes for Factors, Variables, and the Graph.
At the core of the `Factor` template, there is some metaprogramming behind the scenes. Its purpose is converting Ceres function calls to "nice" ones with explicit types. See my detailed notes on "[SLAM] Ceres Interface" board. I have redone this so it does not rely on Boost lirbaries.

It includes one sample Factor type and two sample Variable types, defined in `wave_optimization/tests/factor_graph/example_instances.hpp`. These classes are demonstrated in `wave_optimization/tests/factor_graph/factor_test.cpp`. In `graph_test.cpp` I demonstrate the evaluation of a graph in a trivial triangulation problem. For now, since uncertainty is not implemented, variables themselves are set constant to indicate a "perfect" prior.

Note the `@todo`s appearing in the documentation.


chutsu:
Should I close #57 considering this supersedes it?

wave_optimization/include/wave/optimization/ceres/graph_wrapper.hpp
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @todo change interface to expose results / ceres summary output
+ */
+void evaluateGraph(FactorGraph &graph) {
+    // Initialize the problem
+    FactorGraphProblem problem{graph};
+
+    // Initialize the solver
+    ceres::Solver::Options options;
+    options.linear_solver_type = ceres::DENSE_QR;
```
leokoppel: No, I don't know if we want to use DENSE_QR. This function is mostly a demonstration of using the `FactorGraph` class, not intended be permament.

leokoppel: If we want to expose options to the user we can probably just expose the ceres solver object. No real reason to hide it. I thought the idea was to choose good options, so the user doesn't _need_ to set them unless they really want to tweak it. @chutsu ?

navganti: I figured it was just a demonstration - but I wanted to raise the point.

I do feel like the user *should* be able to set them...or at least a subset of the options that we can specify.

wave_optimization/include/wave/optimization/ceres/graph_wrapper.hpp
```diff
@@ -0,0 +1,83 @@
...
+
+namespace wave {
+/** @addtogroup optimization
+ *  @{ */
+
+class FactorBase;
+class FactorGraph;
+
+
+class FactorGraphProblem : public ceres::Problem {
```
cglwn: Could we avoid this inheritance? I don't think I see this `FactorGraphProblem` class being used polymorphically anywhere, so it might be better to just make `ceres::Problem` a member variable.

wave_optimization/include/wave/optimization/ceres/graph_wrapper.hpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+};
+
+
+/**
+ * Evaluates the factors in graph. Sets the variable to the estimated values.
+ *
+ * @todo change interface to expose results / ceres summary output
+ */
+void evaluateGraph(FactorGraph &graph) {
```
cglwn: Wouldn't this be better placed as a member function of FactorGraph?

wave_optimization/include/wave/optimization/factor_graph/Factor.hpp
```diff
@@ -0,0 +1,138 @@
...
+     * @param[out] jacobians
+     * @parblock
+     * Jacobian matrices with respect to each variable. There is one jacobian
+     * parameter for each variable parameter, in the same order.
+     * @endparblock
+     *
+     * @note  Each residual and jacobian parameter is a special map object which
+     * may evaluate to false to indicate the jacobian should not be calculated
+     * on this call. Wrap calculation and assignment of each parameter `j` in a
+     * conditional, `if (j) {...}`.R
```
cglwn: I'm not sure if I understand the extra `R` at the end of this line. Is that something to do with conditional Jacobian evaluation :thinking:?

wave_optimization/include/wave/optimization/factor_graph/FactorVariable.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * syntactic convenience. It is a data structure which assigns labels to an
+ * underlying value, to simplify interacting with it. For example, a Pose3D view
+ * may define elements of the underlying 6D value as `position` and
+ * `orientation`. In the Factor's evaluation function, these objects can then be
+ * used as independent Eigen vectors.
+ */
+template <typename V>
+class FactorVariable : public FactorVariableBase {
+ public:
+    using ViewType = V;
```
cglwn: Why not use ViewType as the template name? i.e. `template <typename ViewType>`.

Sorry if I'm missing some style guide thing here 😅.

wave_optimization/include/wave/optimization/factor_graph/ValueView.hpp
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * This class is meant to be derived by specialized variable types which assign
+ * meaning to the underlying data vector.
+ *
+ * @todo add more syntactic sugar
+ */
+template <int S>
+class ValueView {
+ public:
+    constexpr static int Size = S;
```
cglwn: Why not just set this as the template variable name. i.e. `template <int Size>`.

Sorry if I'm missing some style guide thing here. 😅

wave_optimization/include/wave/optimization/factor_graph/impl/FactorGraph.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+typename FactorGraph::const_iterator FactorGraph::end() const noexcept {
+    return this->factors.end();
+}
+
+std::ostream &operator<<(std::ostream &os, const FactorGraph &graph) {
+    int index = 0;
+
+    for (auto f : graph.factors) {
+        //        os << "f" << index << " -- ";
```
cglwn: Remove dead code?

wave_optimization/include/wave/optimization/ceres/graph_wrapper.hpp
leokoppel: Yes good point, there is no point to the inheritance.

wave_optimization/include/wave/optimization/factor_graph/Factor.hpp
leokoppel: It is a typo. thanks

wave_optimization/include/wave/optimization/ceres/graph_wrapper.hpp
leokoppel: As long as it uses only public members of FactorGraph is better of as a non-member function.
http://www.drdobbs.com/cpp/how-non-member-functions-improve-encapsu/184401197

But this whole file is up for reorganization.

wave_optimization/include/wave/optimization/factor_graph/FactorVariable.hpp
leokoppel: The template parameter `V` is only in scope inside the template. `using` creates a member typedef name which is actually a member of the class.

cglwn: Ah, I see. I didn't see the type being used anywhere but I see it's being used in the other files now.

wave_optimization/include/wave/optimization/ceres/graph_wrapper.hpp
cglwn: Ok, that rationale makes sense to me.

leokoppel:
Made a few changes after asking @arunxdas about noise handling / normalization.

See commit message

> - Instead of Factor::evaluate, use a free measurement function
> - The measurement function calculates the expected measurement, not residuals
> - A factor has an associated FactorMeasurement holding the actual observed
  value
>
> Thus we can calculate the residual f(x) - Z, outside of "user" code, and later (when noise is implemented) normalize the residual.

leokoppel:
Closing for now, we will get to factor graphs later

PR #143, opened 2017-06-09
Use container-based travis build

leokoppel:
#141 updated to merge into master

The [container-based environment](https://blog.travis-ci.com/2016-11-08-trusty-container-public-beta) has a [faster boot time](https://docs.travis-ci.com/user/ci-environment/). 

It has no sudo, but we don't need it. To install dependencies, use travis' "APT Addon":
    https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-with-the-APT-Addon

The downside is travis would no longer test the dependency install script itself. I think that's not a big loss especially as it uses a different Ubuntu version than most actual users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,30 @@
 dist: trusty
-sudo: required
+sudo: false
 
-language:
-    - c++
+language: cpp
 
 cache:
-    ccache: true
+  ccache: true
 
-before_install:
-    - bash scripts/ci/ci_check.bash
+addons:
+  apt:
+    sources:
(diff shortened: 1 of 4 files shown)


chutsu:
Did you document how you built the PPAs somewhere? This would really benefit the other teams in Wavelab

leokoppel:
Here is packaging documentation: http://packaging.ubuntu.com/html/
For most of the packages I modified an existing Ubuntu package.
Each package on the PPA includes the "debian" source package which has a changelog.

PR #153, opened 2017-06-20
Outlier Rejection

### LINKED ISSUE #145: RANSAC Outlier Rejection
Implement RANSAC outlier rejection to clean up matches returned by BruteForceMatcher.

navganti:
Resolves #145. Implementation of outlier rejection for detected features within the matching interface. 

Includes options to use RANSAC, 7-Point, 8-Point, or least median squares to solve for the fundamental matrix, creating the epipolar constraints. 

Also contains two separate methods as a first filter for matches, either using the ratio test (from  Dr. Lowe's SIFT paper), or a threshold based heuristic. All options are configurable via .yaml files.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


wave_vision/include/wave/vision/matcher/descriptor_matcher.hpp
```diff
@@ -0,0 +1,99 @@
...
+ *  matchers can be found [here][opencv_descriptor_matchers].
+ *
+ *  [opencv_descriptor_matchers]:
+ *  http://docs.opencv.org/trunk/db/d39/classcv_1_1DescriptorMatcher.html
+ */
+
+class DescriptorMatcher {
+ protected:
+    /** Destructor */
+    ~DescriptorMatcher() = default;
```
Jebediah: What does this do?

wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -53,28 +65,133 @@ void BruteForceMatcher::checkConfiguration(
...
+
+    min_distance = matches.begin()->distance;
+
+    // Keep any match that is less than the rejection heuristic times minimum
+    // distance
+    for (auto &match : matches) {
+        if (match.distance <=
+            this->current_config.distance_threshold * min_distance) {
+            filt_matches.push_back(match);
+        }
```
Jebediah: So it will replace the match if it the same distance? Is this desired?

wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -53,28 +65,133 @@ void BruteForceMatcher::checkConfiguration(
...
+        fp1.push_back(keypoints_1.at(match.queryIdx).pt);
+        fp2.push_back(keypoints_2.at(match.trainIdx).pt);
+    }
+
+    // Find fundamental matrix
+    std::vector<uchar> mask;
+    cv::Mat fundamental_matrix;
+
+    fundamental_matrix =
+      cv::findFundamentalMat(fp1, fp2, cv::FM_RANSAC, 3., 0.99, mask);
```
Jebediah: What are these magic numbers?

wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -53,28 +65,133 @@ void BruteForceMatcher::checkConfiguration(
...
+  const cv::Mat &descriptors_2,
+  const std::vector<cv::KeyPoint> &keypoints_1,
+  const std::vector<cv::KeyPoint> &keypoints_2,
+  const cv::InputArray &mask) const {
+    std::vector<cv::DMatch> good_matches;
+    std::vector<cv::DMatch> filtered_matches;
+
+    if (current_config.use_knn) {
+        std::vector<std::vector<cv::DMatch>> matches;
+        int k = 2;
```
Jebediah: magic number

navganti: Yes, so for the ratio test we only want to look at the two nearest neighbours. We don't care about anymore matches, and so this feature was hidden from the user. Is it okay if it's still a magic number?

navganti: okay after discussion with @Jebediah, the decision is to include them in the BFMatcherParams struct, but not include it as a variable in the constructor. Therefore, the user cannot change this.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Fixed. They're parameters to specify the outlier threshold and confidence of the RANSAC estimation. As discussed below, they've been moved into the struct.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Fixed. The user can still change this, but it will take some digging to do that.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: No, so it checks to see if the current match is less than the distance threshold * minimum distance.

If so, it's moved into the new filt_matches vector. The match is never replaced.

wave_vision/include/wave/vision/matcher/descriptor_matcher.hpp
navganti: Using this as reference: https://stackoverflow.com/questions/13576055/how-is-default-different-from-for-default-constructor-and-destructor

It apparently helps differentiate between a _trivial_ and _non-trivial_ class. If I specified the destructor as `~DescriptorMatcher() {}`, it is no longer trivial, due to a user-specified destructor. By stating `~DescriptorMatcher() = default;`, I'm letting the compiler automatically generate the default destructor.

Now perhaps @leokoppel has more to add on my baseline understanding. But as far as I can tell that the main difference has to do with class triviality. My change from `~DescriptorMatcher() {}` to `~DescriptorMatcher() = default;` was a suggestion made in a previous PR.

leokoppel: This class is not trivial anyway; in this case the subtle differences between `{}` and `= default` aren't really important except `default` makes your intentions clear.

The notable thing here is Nav made the destructor `protected` and _non_-`virtual`. This means even though this class is meant to be inherited, derived classes will not be destroyed in an unsafe way (see https://accu.org/index.php/journals/1896).

navganti: Thanks for the clarification @leokoppel :)

wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -5,12 +5,17 @@ namespace wave {
...
+    // Cross_check must be the opposite of use_knn
+    cross_check = !config.use_knn;
+
     // Create cv::BFMatcher object with the desired parameters
     this->brute_force_matcher =
-      cv::BFMatcher::create(config.norm_type, config.cross_check);
+      cv::BFMatcher::create(config.norm_type, cross_check);
 
     // Store configuration parameters within member struct
     this->current_config = config;
```
leokoppel: Do this in initializer list.

Note since you may have a make a copy, the modern approach is to pass by value:
```cpp
BruteForceMatcher(BFMatcherParams config)
  : current_config{std::move(config)} {}
```

https://stackoverflow.com/questions/7592630/is-pass-by-value-a-reasonable-default-in-c11

wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -53,28 +66,140 @@ void BruteForceMatcher::checkConfiguration(
...
+        throw std::invalid_argument(
+          "ratio_threshold is not an appropriate value!");
+    }
+
+    // Check the value of the threshold distance heuristic
+    if (check_config.distance_threshold < 0) {
+        throw std::invalid_argument("distance_threshold is a negative value!");
+    }
+
+    // Only acceptable values are 1, 2, 4, and 8
```
leokoppel: These two checks are a bit confusing. Why not just check explicitly that it's one of `cv::FM_7POINT`, or `cv::FM_8POINT`, etc.

wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -53,28 +66,140 @@ void BruteForceMatcher::checkConfiguration(
...
+        throw std::invalid_argument("fm_method is not an acceptable value!");
+    }
+}
+
+std::vector<cv::DMatch> BruteForceMatcher::filterMatches(
+  std::vector<cv::DMatch> &matches) const {
+    std::vector<cv::DMatch> filt_matches;
+    float min_distance;
+
+    // Sort matches by distance in increasing order
```
leokoppel: Why sort? As far as I see, the sorting is not mentioned in documentation or tested anywhere Even if you don't care about output being sorted, it's an important side effect that this function mutates its input.

If the goal is "get the minimum distance", you don't need to sort.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Yeah, I was considering that after I typed it in. Wasn't sure if this was preferable over doing 4 `or` checks. Will change.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Yeah, goal was to get the minimum distance only. Wasn't sure if I wanted to go through with a for loop...Will fix that and change the function to accept a const input.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: @leokoppel  the reason this wasn't in an initializer list is because I want to check the config for validity before continuing on. If the config is invalid then there is an `std::invalid_argument` thrown from the `checkConfiguration` method.

wave_vision/src/matcher/brute_force_matcher.cpp
leokoppel: You can use [std::min_element](http://en.cppreference.com/w/cpp/algorithm/min_element)

wave_vision/src/matcher/brute_force_matcher.cpp
leokoppel: Ah OK. Theoretically you could make `checkConfiguration` return a reference to it's input, so you could write
```
BruteForceMatcher(BFMatcherParams config)
  : current_config{std::move(checkConfiguration(config))} {}
```

but it's not super important

wave_vision/src/matcher/brute_force_matcher.cpp
leokoppel: The best way would be to use an enum class. Then the value can't possibly be invalid. May add a bit of work wrapping opencv though.

leokoppel: anyway, this is not very important.

navganti: So yes, this is already an enum within openCV. The parser really did not like the enum though...

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Didn't know that existed, thanks @leokoppel :)

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: I feel like at that point we may as well just keep it as is haha

wave_vision/include/wave/vision/matcher/descriptor_matcher.hpp
```diff
@@ -0,0 +1,99 @@
...
+
+ public:
+    /** Matches keypoints descriptors between two images using the
+     *  BruteForceMatcher.
+     *
+     *  @param descriptors_1 the descriptors extracted from the first image.
+     *  @param descriptors_2 the descriptors extracted from the second image.
+     *  @param keypoints_1 the keypoints detected in the first image
+     *  @param keypoints_2 the keypoints detected in the second image
+     *  @param mask The mask variable indicates which descriptors can be matched
```
navganti: Done.

wave_vision/include/wave/vision/matcher/descriptor_matcher.hpp
```diff
@@ -0,0 +1,99 @@
+/**
+ * @file
+ * Base class from which descriptor matchers can be derived.
+ * @ingroup vision
+ */
+#ifndef WAVE_VISION_DESCRIPTOR_MATCHER_HPP
+#define WAVE_VISION_DESCRIPTOR_MATCHER_HPP
+
+/** C++ Headers */
```
navganti: Done.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Fixed.

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: Done.

PR #159, opened 2017-06-22
Feature 147 multithreaded matcher

### LINKED ISSUE #147: Class to multithread matching for batch applications
Paralleling independent matches would speedup the slowest part of the front-end. The solution I have in mind is to create a class that spawns a configurable number of threads each using an instance of a matcher class. External to the class, matches would be added via a member function that blocks until a free thread is available to start matching. 

This class would require the user to keep track of match ids.

Jebediah:
Resolves #147 

Class contains thread pool to speed up batch matches.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


wave_matching/include/wave/matching/impl/multi_matcher_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+template <class T, class R>
+void MultiMatcher<T, R>::initPool(std::string path) {
+    this->config = path;
+    for (int i = 0; i < this->n_thread; i++) {
+        if (this->config.empty()) {
+            this->matchers.push_back(T(R()));
+        } else {
+            this->matchers.push_back(T(R(this->config)));
```
leokoppel: Whenever you write `vector.push_back(T(args...))`, instead write [`vector.emplace_back(args...)`](http://en.cppreference.com/w/cpp/container/vector/emplace)

wave_matching/include/wave/matching/multi_matcher.hpp
```diff
@@ -0,0 +1,83 @@
...
+ * Class is templated for different matcher types
+ * @tparam T matcher type
+ * @tparam R matcher params type
+ */
+template <typename T, typename R>
+class MultiMatcher {
+ public:
+    MultiMatcher(int n_threads = std::thread::hardware_concurrency(),
+                 int queue_s = 10,
+                 std::string path = "")
```
leokoppel: Is there some compulsion to deeply embed filesystem paths into every class?
This class does not need to store or handle file names. Why not accept the params object itself as an argument?

wave_matching/include/wave/matching/impl/multi_matcher_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+        }
+        this->pool.push_back(std::thread(&MultiMatcher<T, R>::spin, this, i));
```
leokoppel: Why does the `MultiMatcher` class store N matcher objects and N threads? Can each thread not construct a matcher itself?

wave_matching/include/wave/matching/multi_matcher.hpp
Jebediah: can't help it. gotta have that sweet sweet filesystem

navganti: @leokoppel is on my case too, @Jebediah. We must stay strong!

wave_matching/include/wave/matching/impl/multi_matcher_impl.hpp
Jebediah: Why does it matter?

leokoppel: Why does it matter that the MultiMatcher class stores N matcher objects and N threads? That was my question! :wink:

You should declare each variable as locally as possible. That is for readability and maintainability. 

As a corollary: if the variables are declared non-locally in some roundabout way, it suggests to readers that your design _requires_ it for some reason. But I can't tell what that is.

Jebediah: Would need another mutex to share the same parameter struct. Kind of annoying.

leokoppel: I don't see how using more local variables could require more mutexes. Let's discuss offline

wave_matching/include/wave/matching/gicp.hpp
```diff
@@ -27,9 +27,18 @@ namespace wave {
...
+    int corr_rand, max_iter;
+    double r_eps, fit_eps;
+    float res;
+};
+
 class GICPMatcher : public Matcher<PCLPointCloud> {
  public:
-    explicit GICPMatcher(float resolution, const std::string &config_path);
+    explicit GICPMatcher(GICPMatcherParams params1);
     /** hey there
```
navganti: hey there

wave_matching/tests/ndt_tests.cpp
```diff
@@ -25,8 +25,9 @@ class NDTTest : public testing::Test {
         pcl::io::loadPCDFile(TEST_SCAN, *(this->ref));
     }
 
-    void initMatcher(const float res, const Affine3 perturb) {
-        this->matcher = new NDTMatcher(res, TEST_CONFIG);
+    void initMatcher(const NDTMatcherParams params,
+                     const Eigen::Affine3d perturb) {
+        this->matcher = new NDTMatcher(params);
```
navganti: why are you manually allocating here? (I know it's a test, sorry. Just curious)

wave_matching/src/gicp.cpp
```diff
@@ -3,35 +3,42 @@
...
-    this->target = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
-    this->final = boost::make_shared<pcl::PointCloud<pcl::PointXYZ> >();
-
-    if (res > 0) {
-        this->resolution = res;
-        this->filter.setLeafSize(res, res, res);
-    } else {
-        this->resolution = -1;
-    }
+GICPMatcherParams::GICPMatcherParams() {
```
navganti: same note regarding defaults

wave_matching/include/wave/matching/impl/multi_matcher_impl.hpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+template <class T, class R>
+void MultiMatcher<T, R>::spin(int threadid) {
+    std::tuple<int, PCLPointCloud, PCLPointCloud> val;
+    while (true) {
```
Jebediah: returning mid loop causes all the mutexes to go out of scope and unlock. cleaner

navganti: Ah I see.

wave_matching/include/wave/matching/gicp.hpp
```diff
@@ -27,9 +27,18 @@ namespace wave {
 /** @addtogroup matching
  *  @{ */
 
+struct GICPMatcherParams {
+    GICPMatcherParams();
+    GICPMatcherParams(const std::string &config_path);
```
Jebediah: That's so manual though.

navganti: Yes but then it will also be consistent with the vision directory :+1:

navganti: (also who knows, could be a useful interface)

wave_matching/include/wave/matching/multi_matcher.hpp
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Class is templated for different matcher types
+ * @tparam T matcher type
+ * @tparam R matcher params type
+ */
+template <typename T, typename R>
+class MultiMatcher {
+ public:
+    MultiMatcher(int n_threads = std::thread::hardware_concurrency(),
+                 int queue_s = 10,
```
Jebediah: Yes, but it is a default argument so it can easily be overidden. Seems like a waste to make a parameter struct for a single parameter.

navganti: Oh sorry, I misread this. For some reason didn't pick up it was in the constructor. Carry on!

wave_matching/tests/ndt_tests.cpp
Jebediah: The constructor sets the configuration so I'm using a pointer so that the configuration can be set later.

PR #170, opened 2017-07-05
VOTestDataset: separate generation and file output

### LINKED ISSUE #169: VOTestDataset writes to a file, not accessible programatically
#### Steps to Reproduce
1. See the test `VOTestDataset.generateTestData` in `wave_vision/tests/dataset_tests.cpp`
(design defect, not bug)

#### Desired behaviour
`VOTestDataset::generateTestData()`  generates a dataset, which I can use in my program.
Writing to a file is done separately.

#### Observed behaviour
`VOTestDataset::generateTestData()` writes to files, without exposing data programatically. (The class does not provide a function to read the files, either).
Simulation math is intermixed with filesystem access.

#### Recommendation
Filesystem access should never be mixed with domain logic.

- If a function accepts a filename as a parameter, that function should _only_ deal with writing or reading a file.
- If a function handles domain logic (e.g., simulates robot motion), it should _not_ know anything about files.

This recommendation supports _modularity_, a key design concept. See http://www.cs.wustl.edu/~levine/courses/cs342/c++/design-principles_4.pdf for one description of modularity.

leokoppel:
Resolves #169 

- Rename VOTestDataset to VOTestDatasetGenerator: a class which generates a dataset
- Add VOTestDataset struct which contains the simulated ground truth and measurements
- Split VOTestDataset::generateTestData() into two functions:
   - VOTestDatasetGenerator::generate() creates the dataset
   - writeTestDataToFile() takes the dataset and writes it to files

#### Pre-Merge Checklist:
- [x] Code has automated tests (only slightly more than before..)
- [x] Zero compiler warnings

### DIFF
--- a/wave_vision/include/wave/vision/dataset.hpp
+++ b/wave_vision/include/wave/vision/dataset.hpp
@@ -63,19 +63,47 @@ class VOTestCamera {
                       std::vector<std::pair<Vec2, Vec3>> &observed);
 };
 
+/** A set of feature observations at one timestep */
+struct VOTestInstant {
+    /** A time in nominal seconds */
+    double time;
+
+    /** The corresponding camera frame, or -1 if no camera observations */
+    int camera_frame = -1;
+
+    /** Robot pose in x, y, z (NWU) */
+    Vec3 robot_pose;
+
+    /** Feature observations where the first of each pair is the measurement in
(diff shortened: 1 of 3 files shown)


chutsu:
BTW just so you know, I made lots of modifications to this class somewhere else. So you might want to avoid making changes to this class.

PR #188, opened 2017-07-24
Fix finding Eigen in CMake

### LINKED ISSUE #186: Fix finding Eigen in CMake
Sub-issue of #45 

The correct way to require Eigen is 

```cmake
FIND_PACKAGE(Eigen3 REQUIRED)
INCLUDE_DIRECTORIES(${EIGEN3_INCLUDE_DIR})
```

leokoppel:
Resolves #186

Replace all uses of absolute paths with correct `FIND_PACKAGE` and `INCLUDE_DIRECTORIES `commands, which work with the `FindEigen3.cmake` we include.

#### Pre-Merge Checklist:
- [x]


wave_vision/CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ ADD_DEFINITIONS(${OpenCV_DEFINITIONS})
 
 # INCLUDES
 INCLUDE_DIRECTORIES(${OpenCV_INCLUDE_DIRS})
-INCLUDE_DIRECTORIES(${EIGEN_INCLUDE_DIR})
+INCLUDE_DIRECTORIES(${EIGEN3_INCLUDE_DIR})
```
navganti: Should these be combined into one `INCLUDE_DIRECTORIES`, like in `wave_utils`?

leokoppel: yes

PR #189, opened 2017-07-24
Fix finding Kindr in CMake

### LINKED ISSUE #187: Fix finding Kindr in CMake
Sub-issue of #45 

The correct way to require Kindr is 

find_package(kindr) 
```cmake
FIND_PACKAGE(kindr REQUIRED)
INCLUDE_DIRECTORIES(${kindr_INCLUDE_DIRS})
```
The current `FindKindr.cmake` should be removed or replaced. Currently it simply lists an absolute path which is not correct for most users. Removal may require some old users to change configuration or reinstall kindr.

leokoppel:
Resolves #187

- Use correct `find_package` command to find kindr
- Remove non-general `FindKindr.cmake`

The removed `FindKindr.cmake` is not a true "Find" script - it is only a renamed copy of the `kindrConfig.cmake` script normally installed by kindr.  It lists an absolute directory which is system-specific, and the particular directory (`/usr/local/src/kindr/kindr-1.0.4`) is atypical.

#### Note
If you have been using libwave since long ago, this removal may break your installation.  You may need to change your system configuration or reinstall kindr (such as by running scripts/install/install_deps.bash)


leokoppel:
Obsolete.
#226

Issue #179, opened 2017-07-17
Add FLANN based matching


navganti:
Add FLANN (Fast Library for Approximate Nearest Neighbors) matching to the repo. This should be faster than BruteForceMatching, hopefully bringing feature matching closer to realtime operation.

PR #197, opened 2017-07-27
Feature 174 ORB Detector + Descriptor

### LINKED ISSUE #174: Interface for ORB Detector/Descriptor
Currently the wave_vision module only includes a FASTDetector and BRISKDescriptor. ORB is proven to provide better information.

navganti:
Resolves #174. Add interface to libwave for the ORB detector and descriptor. Follows the same adaptor pattern as the FASTDetector and BRISKDescriptor developed prior, built off of the FeatureDetector and DescriptorExtractor base classes. Fully tested and commented, and is ready for review ASAP.

The ORB detector and descriptor are extremely quick, and can help facilitate real-time operation. The FLANN matcher would be the last step to allowing truly real-time operation; that is discussed in #179.

It probably should be done in other PRs, but I've made slight modifications to the FAST and BRISK tests, primarily to make them more robust and also follow the same pattern as the ORB Tests.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


navganti:
Here are tracking results with the ORB detector + descriptor. There are less outliers than the FASTDetector results.

However, one parameter of the ORBDetector is the number of features to retain. I've set the default for this to be 2000, and so after outlier rejection there are a smaller number of matches left. In comparison, the FASTDetector initially detects ~4500 features. However, this is completely configurable, and we can make the ORBDetector retain more features. In short, while there are less outliers, there are also less keypoints to start with.

[feature_tracks_orb.zip](https://github.com/wavelab/libwave/files/1181371/feature_tracks_orb.zip)

wave_vision/include/wave/vision/descriptor/orb_descriptor.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    /** The size of the square patch used in the random point sampling to
+     *  construct the descriptor. This patch is smoothed using an integral
+     *  image.
+     *
+     *  Default: 31 (31 x 31 pixels). Must be greater than 0.
+     */
+    int patch_size = 31;
+
+    // Default parameters that should not be modified
```
leokoppel: A bit confusing; why do you have them if they should not be modified? Does OpenCV us them in the detector but not the descriptor? If so, maybe they don't belong here at all?

wave_vision/tests/viz_tests/tracker_viz_tests.cpp
```diff
@@ -45,10 +47,48 @@ TEST(TrackerTests, OfflineTrackerTest) {
...
+    ORBDescriptor descriptor;
+    BruteForceMatcher matcher;
+
+    Tracker<ORBDetector, ORBDescriptor, BruteForceMatcher> tracker(
+      detector, descriptor, matcher);
+
+    image_sequence = readImageSequence(FIRST_IMG_PATH);
+
+    feature_tracks = tracker.offlineTracker(image_sequence);
+    ASSERT_NE((int) feature_tracks.size(), 0);
```
leokoppel: I like to use `u` suffix to avoid casts in these checks. E.g.  `ASSERT_NE(0u, feature_tracks.size())`;

Here could also do `ASSERT_FALSE(feature_tracks.empty())`

wave_vision/tests/detector_tests/orb_tests.cpp
```diff
@@ -0,0 +1,137 @@
...
+    ORBDetectorParams config;
+    config.fast_threshold = 0;
+
+    ASSERT_THROW(ORBDetector bad_fastthreshold(config), std::invalid_argument);
+
+    config.fast_threshold = -1;
+    ASSERT_THROW(ORBDetector bad_fastthreshold1(config), std::invalid_argument);
+}
+
+TEST(ORBDetectorTests, ConfigurationTests) {
```
leokoppel: These are some very detailed tests of the configuration functionality, :hushed: :+1:

wave_vision/tests/config/descriptor/orb.yaml
```diff
@@ -0,0 +1,24 @@
...
+# output result will be 2 bits, as the result is the index of the
+# brightest point. Therefore, the distance method used for the
+# BruteForceMatcher or FLANNMatcher will have to be cv::NORM_HAMMING_2, a
+# variation to the Hamming distance, when WTA_K = 3 or 4.
+#
+# Options: 2, 3, or 4.
+#
+# Default: 2.
+
+wta_k: 2
```
leokoppel: I don't know if it's better to leave the names exactly as in OpenCV (so people can refer to their documentation), or change to something more descriptive than `wta_k`

wave_vision/include/wave/vision/descriptor/orb_descriptor.hpp
navganti: Yeah, exactly that @leokoppel . The thing is that they're required when I call the `cv::ORB::create()` method in the `ORBDescriptor` constructor. The detector has the opposite...only `wta_k` and `patch_size` are defined as "do not touch".

I don't think I want to keep these values defined in the `ORBDescriptor` constructor...should I just make these `Params` structs into classes and protect them? They have constructor methods, so making them classes wouldn't be the silliest idea.

wave_vision/tests/config/descriptor/orb.yaml
navganti: I had the exact same thoughts. `wta_k` is so uninformative. I'll change it to something better.

wave_vision/tests/viz_tests/tracker_viz_tests.cpp
navganti: Good call. I like `ASSERT_FALSE`.

navganti: fixed.

wave_vision/include/wave/vision/descriptor/orb_descriptor.hpp
navganti: @leokoppel and I talked in person, these have been moved into the constructor.

wave_vision/tests/config/descriptor/orb.yaml
navganti: Okay, renamed to `tuple_size`. It's a bit better...I got the name from [here](https://books.google.ca/books?id=LPm3DQAAQBAJ&pg=PA564&lpg=PA564&dq=ORB+WTA_K+meaning&source=bl&ots=2vHjSa7iyb&sig=Oxva0bFdtSApDtBNwNqSND79h5Y&hl=en&sa=X&ved=0ahUKEwjs5JGt2arVAhXHFT4KHVH-AmoQ6AEIXTAI#v=onepage&q=ORB%20WTA_K%20meaning&f=false).

PR #200, opened 2017-07-30
Enhance 198 matcher diagnostics

### LINKED ISSUE #198: Add some diagnostic information to feature_matcher
In order to evaluate the performance of the tracker and matcher, one piece of useful information is the number of features rejected by the RANSAC classifier. This information is currently tucked away, and so the matcher may need to be slightly restructured.

navganti:
Resolves #198. In order to access diagnostic information, the removeOutliers method is no longer automatically called unless specified by the user. Useful  diagnostic information includes the amount of matches filtered by the outlier removal, and this change allows access to that info. This also provides more flexibility for the user, as they can define their own outlier rejection method if they choose.

Tests have also been made more robust, and are uniform with other tests in the wave_vsion suite.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


wave_vision/src/matcher/brute_force_matcher.cpp
```diff
@@ -84,41 +85,38 @@ void BruteForceMatcher::checkConfiguration(
...
-    std::vector<cv::DMatch> filt_matches;
-    float min_distance;
-    std::vector<cv::DMatch>::iterator closest_match;
+    std::vector<cv::DMatch> filtered_matches;
 
     // Determine closest match
-    closest_match = std::min_element(matches.begin(), matches.end());
-    min_distance = closest_match->distance;
+    auto closest_match = std::min_element(matches.begin(), matches.end());
+    float min_distance = closest_match->distance;
```
leokoppel: `auto` seems right here to avoid any future troubles e.g. if this becomes a `double`

wave_vision/include/wave/vision/matcher/descriptor_matcher.hpp
```diff
@@ -89,7 +89,7 @@ class DescriptorMatcher {
       const cv::Mat &descriptors_2,
       const std::vector<cv::KeyPoint> &keypoints_1,
       const std::vector<cv::KeyPoint> &keypoints_2,
-      cv::InputArray mask = cv::noArray()) const = 0;
+      cv::InputArray mask) const = 0;
```
leokoppel: Update doxygen

wave_vision/src/matcher/brute_force_matcher.cpp
navganti: I probably need to do an `auto` run through of all the code. Maybe part of a future cleanup.

wave_vision/include/wave/vision/matcher/descriptor_matcher.hpp
navganti: @leokoppel this was actually the place where I mentioned CLion was complaining about overriding the default parameters of a virtual function, btw. It says "default arguments on virtual or override methods are prohibited." It's still yelling at me, actually.

navganti: Any suggestions? I'm not sure what the best way to do this is. The default should be no mask. 

I want to say these classes will not be used in a polymorphic manner...but that's always a dangerous assumption. I'm going to add back the default parameter in the base class for now.

leokoppel: A foolproof way is to make a function with the same name that takes 4 arguments. Have it call the version with 5 arguments.

navganti: Nice. I like that. Will add that fix in here.

PR #207, opened 2017-08-08
Increase GCC version to 5.4

### LINKED ISSUE #150: Bugs in gcc 4.8
Platform: Ubuntu 14.04 (as on Travis CI)

gcc 4.8, the default on Travis, has bugs in its C++11 implementation which require our code to be clumsier to work around them. For example, [Constructor inheritance with non-default constructible members](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67054)

gcc 4.8 also implements an old version of the standard. There are few cases where we ran into defects in C++11 which were changed (e.g. [List-initialization of aggregate from same-type object](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#1467))

(No issue links because these build errors had to be fixed before going into master).

#### Suggestion
Increase the required version of gcc, and install non-default gcc on travis.
Of course, we'd lose support for the default gcc on Ubuntu 14.04.

leokoppel:
Resolves #150

Increase GCC requirement and travis version to 5.4.

I suspect 5.0 or 5.1 would work but that is harder to test. 5.4 is the version on Ubuntu 16.04.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,21 @@
 dist: trusty
 sudo: false
 
-language: cpp
+language: generic  # don't use the default compilers for "cpp"
 
 cache:
   ccache: true
 
+env:
+  CXX='ccache g++-5' CC='ccache gcc-5'
+
 addons:
   apt:
     sources:
+      - ubuntu-toolchain-r-test
       - sourceline: 'ppa:lkoppel/robotics'
     packages:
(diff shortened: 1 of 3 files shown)


chutsu:
lgtm :+1:

PR #148, opened 2017-06-15
Add camera models from kalibr

leokoppel:
Resolves #72 

Add camera models from kalibr's aslam_cv/aslam_cameras to deps/aslam_cameras

My process was to copy everything in aslam_cameras and remove things we don't need
that were non-trivial to compile. I had to add a few utility files that were in kalibr's sm ("Schweizer-Messer" == Swiss Army Knife) library. The goal was to keep the actual camera model functions mostly untouched, but not have to include all of kalibr as a dependency.

As such there may be some unneeded functions, but I don't know fully how much will be useful yet.

I removed everything related to:
- serialization
- calibration targets
- constructing from PropertyTree (confuguration object)
- landmarks, descriptors, and frames
- some versions of methods outputting uncertainty

I changed:
- custom `Duration` class in rolling shutter model to `std::chrono::duration` (it's not thoroughly tested, but we don't use this yet)

Testing: kalibr's unittests

I did not clang-format this code. I see it as a third-party inclusion (note **the way to `include` these headers from our other code is no different than if all of kalibr were installed as a dependency**). Keeping original formatting may help with things such as `diff`ing it against the original source.

**Pros**
- Fully featured camera models (pinhole, omni, distortions, with jacobians, etc)
- Took a day of work to add, instead of much longer
- No external dependency users must install

**Cons**
- "Original" BSD license with advertising clause
- Does not follow our formatting, style, or documentation standards
- We essentially maintain a fork of the dependency

#### Pre-Merge Checklist:
- [ ] Code is documented in doxygen format
- [x] Code has automated tests
- [x] Zero compiler warnings
- [ ] Formatted with `clang-format`

PR #214, opened 2017-08-11
Feature 211 Kalibr AprilTags

### LINKED ISSUE #211: Integrate AprilGrid detection from Kalibr into libwave
Looking to incorporate AprilGrid detection code from Kalibr into libwave. This will help our calibration efforts.

navganti:
Resolves #211. As per a request from @arunxdas, this brings in Kalibr's April tags library, as well as anything it depends on. My attempt at doing this in is in a similar manner to #148 - I copied over the main AprilTagsDetection code, and then anything it depended on. There may be a few functions within the april tags library that can be removed.

I used Kalibr's unittests for the functions that are used by the AprilTags library. Unlike #148, the tests are not as extensive. 

No doxygen, no clang-format. As @leokoppel stated in his PR, we're essentially maintaining a fork of Kalibr's library. This is also using their original BSD licence. 

I made no modifications to any of their code.

#### Pre-Merge Checklist:
- [] Code is documented in doxygen format
- [] Zero compiler warnings
- [] Formatted with `clang-format`


navganti:
I believe @arunxdas was looking to develop an app that used this...I suggest prior to merging this in, the app can verify that this library operates as intended.

Since there are no actual tests for the AprilTag detection, perhaps the app can act as that.

navganti:
Also @leokoppel @chutsu  this is failing on a `-Werror=unknown-pragmas`. Should I go into the code and work around this? Or can this just be fixed via a flag or something?

leokoppel:
You need extra flags to use openmp, normally through `find_package(OpenMP)` -- but it looks like kalibr's cmake configuration doesn't actually do that, and it doesn't enable warnings either.

I would just comment out the `#pragma` line.

navganti:
Sounds good, thanks @leokoppel. Also - most of this code was within the `aslam_cameras_april` package within Kalibr, except the `apriltags` library itself was within the `aslam_offline_calibration` package. I combined this...should I have brought over the offline calibration package was well?

deps/aslam_cameras_april/test/logging/logTest.cpp
```diff
@@ -0,0 +1,49 @@
+
```
leokoppel: Is this test run? Do we really need it? (same for the other similar files)

navganti: I included it because it's the Kalibr test for a dependency of the AprilTags library. It looks like the `TagDetector` or one of those similar files calls the Logger, and this is the test for that.

I believe the test was run? I ran it on my computer anyways, maybe I set it up wrong

navganti: Yup, it is run! There were some other tests that I didn't include from this folder, as the files didn't include the relevant headers.

navganti:
@leokoppel that sounds like a good idea. @arunxdas, definitely follow that process when developing your app!

navganti:
Closing this PR for now, there are a few things that need to happen before this is merged in. Currently, the deps are exposed for @arunxdas to develop his app in `libwave_apps`, but this should be incorporated into `wave_vision` eventually.

PR #216, opened 2017-08-14
Front End Evaluation Tools

### LINKED ISSUE #215: front end evaluation tools
This builds on #198 and #200. Turns out more information had to be exposed in order to get all the diagnostic information we wanted. This includes some values regarding number of keypoints detected, filtered, good, and raw matches, etc.

navganti:
Resolves #215. Exposes a bit more information from the tracker, keypoint detectors, and matchers for diagnostics info.

Also incorporates some other features. I've added in a `num_features` that can be retained on the FAST detector. I've also deleted all of that unnecessary data from the KITTI dataset, and the associated tests.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


navganti:
I've also now added in a "sliding window" tracker operation mode, which will throw out measurements from the measurement container that are in frames older than a configurable amount.

 I found that this greatly improved performance of the front end - at least when trying to extract/draw tracks at every frame.

navganti:
Hi @Jebediah @leokoppel @chutsu - long time since this was opened, but could one of you guys review this if you have a second?

wave_vision/include/wave/vision/detector/fast_detector.hpp
```diff
@@ -68,6 +81,9 @@ struct FASTDetectorParams {
  */
 class FASTDetector : public FeatureDetector {
  public:
+    /** The number of keypoints detected. */
+    size_t num_keypoints = 0u;
```
chutsu: Put this into FeatureDetector class?

wave_vision/include/wave/vision/matcher/flann_matcher.hpp
```diff
@@ -179,6 +179,20 @@ struct FLANNMatcherParams {
...
+
+    /** The number of matches after being filtered by the distance/ratio test.
+     *  This is the input to the outlierRejection method.
+     */
+    size_t num_filtered_matches = 0u;
+
+    /** The size of matches remaining after outlier removal via RANSAC or a
+     *  similar method.
+     */
+    size_t num_good_matches = 0u;
```
chutsu: Put these into the DescriptorMatcher class?

wave_vision/include/wave/vision/matcher/flann_matcher.hpp
```diff
@@ -231,7 +245,7 @@ class FLANNMatcher : public DescriptorMatcher {
       cv::Mat &descriptors_2,
       const std::vector<cv::KeyPoint> &keypoints_1,
       const std::vector<cv::KeyPoint> &keypoints_2,
-      cv::InputArray mask = cv::noArray()) const override;
+      cv::InputArray mask = cv::noArray()) override;
```
chutsu: Maybe add virtual at the beginning of the function declartion?

wave_vision/include/wave/vision/tracker/tracker.hpp
```diff
@@ -50,7 +60,7 @@ class Tracker {
      * @return tracks corresponding to all detected landmarks in the image, from
      * the start of time to the given image.
      */
-    std::vector<FeatureTrack> getTracks(const size_t &img_num) const;
+    std::vector<FeatureTrack> getTracks(const int img_num) const;
```
chutsu: int?

wave_vision/include/wave/vision/detector/fast_detector.hpp
navganti: done, thank you

wave_vision/include/wave/vision/matcher/flann_matcher.hpp
navganti: fixed, thank you.

wave_vision/include/wave/vision/matcher/flann_matcher.hpp
navganti: okay I did some searching. I don't think this is necessary. The virtual keyword I've used in the base class, and now I'm overriding it here.

It looks like you can write virtual and there is no change in behaviour but I feel like that just adds confusion?

wave_vision/include/wave/vision/tracker/tracker.hpp
navganti: changed. I can't really remember why I changed it. I think there will be some significant changes coming here as we start using this again.

wave_vision/include/wave/vision/descriptor/brisk_descriptor.hpp
```diff
@@ -23,7 +23,7 @@ namespace wave {
  *  brightness comparison tests to construct the descriptor.
  */
 struct BRISKDescriptorParams {
-    BRISKDescriptorParams() {}
+    BRISKDescriptorParams() = default;
 
     BRISKDescriptorParams(std::vector<float> rlist,
                           std::vector<int> nlist,
```
chutsu: const?

wave_vision/include/wave/vision/descriptor/orb_descriptor.hpp
```diff
@@ -27,7 +27,7 @@ namespace wave {
  *  http://docs.opencv.org/trunk/db/d95/classcv_1_1ORB.html#adc371099dc902a9674bd98936e79739c
  */
 struct ORBDescriptorParams {
-    ORBDescriptorParams() {}
+    ORBDescriptorParams() = default;
 
     ORBDescriptorParams(int tuple_size, int patch_size)
```
chutsu: const?

wave_vision/include/wave/vision/detector/fast_detector.hpp
```diff
@@ -18,18 +18,22 @@ namespace wave {
...
 /** Configuration parameters for the FASTDetector. */
 struct FASTDetectorParams {
-    FASTDetectorParams() {}
+    FASTDetectorParams() = default;
 
-    FASTDetectorParams(int threshold, bool nonmax_suppression, int type)
+    FASTDetectorParams(int threshold,
+                       bool nonmax_suppression,
+                       int type,
+                       int num_features)
```
chutsu: const?

navganti: But I'm changing the member variables in the initializer list. Wouldn't that not be const?

wave_vision/include/wave/vision/matcher/flann_matcher.hpp
chutsu: virtual in the derived class is purely stylistic.just more explicit.

wave_vision/include/wave/vision/detector/fast_detector.hpp
chutsu: I do not follow. You're passing by value, if we follow the "const" correct way, everything should be const if the function argument is not going to change. Which in this case threshold, nonmax_suppression, type... do not. They are passed by value to member variables.

navganti: OH shoot yes I see, I thought you were talking about the method itself, not the arguments. thank you!

chutsu: It really isn't a big deal with these, because these are primitives, and you can only pass by value, there is no way the variables you pass in would be changed so `const` is not really necessary.

But! IMO, if we are to follow `const` correct-ness we should put `const` infront of primitives anyways, just so that you don't have to think much.

I find this SO answer provides a good rationale why even primitives should have `const`.
https://stackoverflow.com/questions/16033751/c-const-modifier-with-primitive-types

Issue #139, opened 2017-06-08
Ensure correct alignment for Eigen objects


leokoppel:
#### Platform
All

#### Observed result
No observed errors for us yet, but tales of such errors circulate 
E.g. https://github.com/JakobEngel/dso/issues/ 22 

#### Required tasks
- Read about Eigen [alignment issues](https://eigen.tuxfamily.org/dox/group__DenseMatrixManipulation__Alignement.html)
- Find violations in our code and fix them.

#### Related issues
#138

PR #218, opened 2017-10-13
Fix Eigen alignment issue with Matcher

leokoppel:
I was seeing a sefault in MultiTests.initialization, which went away with this change.

See https://eigen.tuxfamily.org/dox/group__DenseMatrixManipulation__Alignement.html

Note I only fixed one instance that was causing a crash; more changes must be made elsewhere. Related: #139


leokoppel:
Yes, and whenever such a struct is stored in an STL container it needs the `Eigen::aligned_allocator`.

PR #222, opened 2017-10-18
Add FindBoost module from latest CMake

leokoppel:
Add FindBoost from CMake 3.9.4

This script defines handy imported targets such as Boost::filesystem. They are
otherwise only bundled with CMake 3.5 and above, which we don't want to require
for now.

(They will be used in a later PR)

Testing: ran CMake and built on my system (CMake 3.5)


navganti:
Will this break anything for users using a lower version of CMake? We maybe should just upgrade the minimum required...we can figure that out on our other projects as we move forward.

leokoppel:
Yes you are right. This module doesn't actually work on 2.8 anyway.

Issue #45, opened 2017-04-13
Properly find dependencies with CMake


leokoppel:
The purpose of CMake is to allow cross-platform builds without making assumptions about directory structure.

Over time some absolute paths have crept into CMakeLists files. Conversely, some dependencies are found by virtue of being in the default path, but are never declared. These practices may cause problems for users with different directory structures, different OS, multiple versions of libraries installed, etc.

The tasks for this issue are:

#### Remove hardcoded paths from CMake files
   This includes lines such as
   ```cmake
   INCLUDE_DIRECTORIES(/usr/include/eigen3)
   ```
   and 
   ```cmake
   INCLUDE_DIRECTORIES(/usr/include)
   ```

#### Explicitly list all dependencies
   For example, Boost is not listed

leokoppel:
In case its useful, here are my slides on where `find_package` looks for things: https://docs.google.com/presentation/d/1zyuT8qPTHDNuhl5jSF2-NTLZ8PPnuNF0VHDxaFFHuPM/edit?usp=sharing

jitrc:
Yes and I use this for Eigen
```
find_package(Eigen3 QUIET)
if(NOT EIGEN3_FOUND)
  # Fallback to cmake_modules
  find_package(cmake_modules REQUIRED)
  find_package(Eigen REQUIRED)
  set(EIGEN3_INCLUDE_DIRS ${EIGEN_INCLUDE_DIRS})
  set(EIGEN3_LIBRARIES ${EIGEN_LIBRARIES})  # Not strictly necessary as Eigen is head only
  # Possibly map additional variables to the EIGEN3_ prefix.
else()
  set(EIGEN3_INCLUDE_DIRS ${EIGEN3_INCLUDE_DIR})
endif()
```

chutsu:
Sure go ahead, there are already 2 (Kindr and Eigen cmake `Find` files in the `libwave/cmake` dir), but fully agree on listing all dependencies

chutsu:
Consensus has been made, we should use cmake files to find dependencies, closing this now but should create new issues for independent issues for each new cmake dependency config file, e.g. PCL, Ceres.

leokoppel:
This issue is a bit broad but it's pretty common to have an overarching issue which is mentioned in sub-issues or multiple PRs. I see it as still open.

Note the issue description says nothing about the creation of "cmake dependency config files". That is not inherent to the issue.

> leokoppel added a commit that referenced this issue 9 hours ago

[...]

PR #223, opened 2017-10-18
Properly link Eigen3 in CMake

leokoppel:
Part of #45 
This is a combination of 4 commits.

#### Add modern Eigen3 imported target

See comments in FindEigen3.cmake

#### Move FIND_PACKAGE(Eigen3) to top-level

It's used by all packages, so should be included in one place to ensure version
is consistent.

#### Link to Eigen target, instead of INCLUDE_DIRECTORIES

This method is more explicit and reliable.

#### Link wave_containers to Eigen

Need to define a library for wave_containers although it's header only.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings
- [ ] Formatted with `clang-format`


navganti:
Add modern Eigen3 imported target

Move FIND_PACKAGE(Eigen3) to top-level

Link to Eigen target, instead of INCLUDE_DIRECTORIES

Link wave_containers to Eigen

PR #225, opened 2017-10-19
Refactor CMake dependencies

### LINKED ISSUE #45: Properly find dependencies with CMake

leokoppel:
Resolves #45

Link against _imported targets_ and set _properties_ instead of using INCLUDE_DIRECTORIES and the like. This method is more explicit, and makes dependencies part of the _link interface_ of each module. 

See https://rix0r.nl/blog/2015/08/13/cmake-guide/

For example: if I link a target to wave_utils, yaml-cpp is automatically linked. Also, wave_utils/include is automatically added to  `INCLUDE_DIRECTORIES` for that target.

There is no more need to `INCLUDE_DIRECTORIES` everything at the top level, or to explicitly call `ADD_DEPENDENCIES` at the top level.


wave_matching/CMakeLists.txt
```diff
@@ -3,38 +3,56 @@ PROJECT(wave_matching)
...
+    ${PCL_SEARCH_LIBRARY}
+    ${PCL_IO_LIBRARY}
+    ${PCL_VISUALIZATION_LIBRARY}
+    ${BOOST_LIBRARIES}
+    ${VTK_LIBRARIES})
 
-# LIBRARY DIRS
+# PCL adds some entries with spaces instead of as a (semicolon-separated) CMake
+# list, which messes up later. Fix them here.
+string(REPLACE " " ";" pcl_definitions "${PCL_DEFINITIONS}")
```
navganti: @leokoppel you just told me you want to fix this. I'm not sure if this counts as my review anymore.

wave_vision/CMakeLists.txt
```diff
@@ -32,7 +17,19 @@ ADD_LIBRARY(
...
+TARGET_INCLUDE_DIRECTORIES(${PROJECT_NAME} PUBLIC include)
+
+# Depend on these modules and external libraries, and make clients use them
+TARGET_LINK_LIBRARIES(${PROJECT_NAME} PUBLIC
+    wave_kinematics
+    wave_utils
+    wave_containers
+    Eigen3::Eigen
+    Boost::filesystem
+    ${OpenCV_LIBS})
```
navganti: is it ok to keep this as OpenCV_libs? Should this be updated to reflect the components that are linked? or is too many components?

leokoppel: `OpenCV_libs` only includes the requested components

navganti: cool. But you changed it above in utils to just include opencv_core?

leokoppel: Yes, utils only uses opencv_core.
In next PR I move all find_package to the top level anyway

wave_matching/CMakeLists.txt
leokoppel: Done, you can review

cmake/FindPCL.cmake
```diff
@@ -0,0 +1,72 @@
...
+    FOREACH(component ${PCL_FIND_COMPONENTS})
+        STRING(TOUPPER "${component}" component)
+        LIST(APPEND imported_pcl_libraries ${PCL_${component}_LIBRARY})
+
+        # Allow these out of the scope
+        SET(PCL_${component}_FOUND ${PCL_${component}_FOUND} PARENT_SCOPE)
+        SET(PCL_${component}_LIBRARY ${PCL_${component}_LIBRARY} PARENT_SCOPE)
+    ENDFOREACH()
+
+    # PCLConfig.cmake is evil for setting these
```
navganti: :joy:

wave_geometry/CMakeLists.txt
```diff
@@ -1,22 +1,26 @@
 PROJECT(wave_geometry)
 
 # EXTERNAL PACKAGES
-FIND_PACKAGE(kindr)
+FIND_PACKAGE(kindr REQUIRED)
```
navganti: should Kindr be capitalized here?

wave_kinematics/CMakeLists.txt
```diff
@@ -23,9 +23,4 @@ WAVE_ADD_TEST(
     tests/quadrotor_test.cpp
     tests/two_wheel_test.cpp
 )
-TARGET_LINK_LIBRARIES(
-    ${PROJECT_NAME}_tests
-    ${PROJECT_NAME}
-    wave_controls
-    wave_utils
-)
+TARGET_LINK_LIBRARIES(${PROJECT_NAME}_tests ${PROJECT_NAME})
```
navganti: much cleaner with clients being able to access the previously linked libraries too

wave_geometry/CMakeLists.txt
leokoppel: No, it must match the name of `kindrConfig.cmake`

leokoppel:
This is half of the work of making modern CMake targets: each wave library is correctly linked to its own dependencies. The next PR will fix the yaml-cpp issue by properly exporting the wave libraries to users.

PR #226, opened 2017-10-22
Overhaul CMake configuration

### LINKED ISSUE #187: Fix finding Kindr in CMake

### LINKED ISSUE #201: CMake config needed for using libwave as a dependency
When trying to use libwave as a dependency for another package, other packages (that libwave already depends on) must be included in the CMake file for it to build properly. The two that I have noticed in particular are `Eigen` and `yaml-cpp`.

Both must be included in your CMakeLists.txt using `find_package` and `include_directories`. `yaml-cpp` must ALSO be included in your `target_link_libraries`, on a line AFTER `wave`.

I'm not sure if this is the expected behaviour. I think that I would expect libwave to be a standalone library that, so long as you have the dependencies installed your computer, can be linked as is.

Below are the errors from not including Eigen, and not including yaml-cpp.

eigen_error

yaml_error

leokoppel:
Resolves #201 
Resolves #187

The major changes are:

- Allow `find_package(wave)` in CMake
- Do not build combined `libwave.a`/`.so` anymore; instead, install all of the component libraries
    - no more "malformed archive" errors
- To link all libwave, use `target_link_libraries(my_target wave::wave)` . To link individual components, use `target_link_libraries(my_target wave::geometry)`, etc
- The above commands link everything: libraries, includes, compile flags, transitive dependencies. There is no need for `include_directories()` or `find_package(yaml-cpp)` in client code.

See [The Ultimate Guide to Modern CMake](https://rix0r.nl/blog/2015/08/13/cmake-guide/) for an intro to what I am going for, though be aware it not really "ultimate".
See docs on [Imported Targets](https://cmake.org/cmake/help/v3.5/manual/cmake-buildsystem.7.html#imported-targets)

To do this, we:
- Only link to `IMPORTED` targets.
   - Where our dependencies do not provide `IMPORTED` targets, we have to define them.
   - This is done in the `Import*` scripts. (Needed for PCL and kindr, and older versions of Eigen and Boost. OpenCV, ceres, and yaml-cpp already provide imported targets, though they don't have `::` syntax).
   - Move all `find_package` to top-level CMakeLists so they can be defined in one place
- Make sure everything we use to build libwave is stored in the correct _properties_ of each target, so anyone who links against libwave automatically gets that information. Install targets and includes in correct destination.
   - Add a few helper functions to do this consistently in each module: `WAVE_ADD_LIBRARY` and 
`WAVE_INCLUDE_DIRECTORIES`
   - Note each subdirectory defines its own target and install destinations (helped by the helper functions) . The top-level CMakeLists does not set target properties.
- _Export_ our own targets so other packages can use them
   - Put them in the `wave::` namespace, e.g. `wave::utils`. The colons `::` are used so that CMake knows it is not a filename and checks that the target exists.
- Generate  `waveConfig.cmake`, `waveConfigVersion.cmake` and `waveTargets.cmake`. These files are used when `find_package(wave)` is called.
- In `waveConfig`, find our dependencies and imported targets again on the host system. 
    - Must include `Import*` scripts with the installation

Changes _not_ made:
- Although each library is exported and aliased as `wave::utils`, it is still defined internally as `wave_utils`. This could be changed to just `utils` (which seems too generic) and the directories could be renamed as well (which seems disruptive).
- Defining components (allowing `find_package(wave COMPONENTS geometry)`) is a nice-to-have for the future.
- Making some components optional (e.g. not building `wave_matching` if PCL is not installed) is a nice-to-have for the future.


cmake/WaveUtils.cmake
```diff
@@ -1,6 +1,7 @@
 # - Macros used in libwave
 
 INCLUDE(CMakeParseArguments)
```
navganti: Super nitpicky - but should this file be called WaveUtils.cmake? Could it be named WaveCMakeUtils.cmake? 

Just because we have a wave_utils package. Idk, just personal preference

CMakeLists.txt
```diff
@@ -1,88 +1,127 @@
...
+# Add an export file listing all targets
+EXPORT(EXPORT WaveTargets NAMESPACE Wave:: FILE WaveTargets.cmake)
+# Install the export file
+INSTALL(EXPORT WaveTargets NAMESPACE Wave:: DESTINATION "${INSTALL_CMAKE_DIR}")
+
+# Create the WaveConfig and WaveConfigVersion files
+
+# Set variables used to generate the Config file
+# Using these is "not recommended". For example, WAVE_INCLUDE_DIRS is not
+# relocatable; WAVE_LIBRARIES gives names but not paths
```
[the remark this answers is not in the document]
leokoppel: Recommended method is described in README.

Yeah maybe I should just take these out; not much point in introducing something that is immediately deprecated.

cmake/WaveUtils.cmake
leokoppel: Good point, will rename WaveHelpers or something. Maybe in subsequent PR or history will be messed up.

leokoppel:
> As we move forward and new versions of PCL/OpenCV/etc are released, how much modification do you think these files will require?

Most of the scripts are to fill in gaps with older versions, and are already not needed for the newest versions. To illustrate, here are possible decisions in the future:

- We require Eigen >= 3.3.1: can remove FindEigen and ImportEigen script.
- We require CMake >= 3.5: can remove ImportBoost script

PCL is the unknown. Probably in some future version they will redo their config file, which might require some changes. Just like any non-backward compatible change to a dependency.

navganti:
Okay, I am on board with that.

leokoppel:
@chutsu made the changes you requested. The package is now all-lower-case `wave`.

Issue #227, opened 2017-10-23
Measurement container performance scales poorly


leokoppel:
An informal report from @navganti 
#### Platform
xenial 64

#### Steps to Reproduce
1. Add items to landmark measurement containers

#### Observed result
Operations become noticeably slow unless old items are deleted from the container.

Should add a benchmark first to determine which operations are slow, then determine how to fix.

leokoppel:
The benchmarks in #228 reveal MeasurementContainer::get() is terribly inefficient: O(NlogN) instead of expected O(logN).

PR #228, opened 2017-10-26
Add benchmarks to MeasurementContainer

leokoppel:
Part of #227 

- Add CMake helpers to add benchmarks using Google's benchmark library
- Add benchmarks investigate MeasurementContainer performance (empalce() and get())

To build them, you must install [benchmark](https://github.com/google/benchmark/), then build libwave with `cmake .. -DBUILD_BENCHMARKS=ON`. 

benchmark is similar to gtest in how the tests are defined; that's part of why I chose it.

#### Pre-Merge Checklist:
- [ ] Code has automated tests
- [x] Zero compiler warnings

These benchmarks are currently not built on Travis. They require an extra library and take pretty long to run. Conceivably in the future we could have a separate benchmarking build on Travis, but it's not a priority.

### DIFF
--- a/cmake/WaveUtils.cmake
+++ b/cmake/WaveUtils.cmake
@@ -40,6 +40,24 @@ FUNCTION(WAVE_ADD_TEST NAME)
 
 ENDFUNCTION(WAVE_ADD_TEST)
 
+# wave_add_benchmark: Add a target which links against google benchmark
+#
+# WAVE_ADD_BENCHMARK(Name src1 [src2...])
+#
+# The target will be added to the tests run by "make benchmark". It will be
+# linked against the needed libraries. Any other links can be made separately
+# with the target_link_libraries command.
+FUNCTION(WAVE_ADD_BENCHMARK NAME)
+    # Build the executable using the given sources
+    ADD_EXECUTABLE(${NAME} ${ARGN})
+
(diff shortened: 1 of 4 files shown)


navganti:
@leokoppel, why does the benchmark need to be added to libwave? It seems to me like it's a tool that you used to evaluate your containers (and would be useful for any code that we develop). Why does it need to be in the actual repo?

leokoppel:
Oh, I left an incorrect comment in CMakeLists.

`benchmark`, the library is not in the repo and is not built when building libwave. (though that could change, just like we do build gtest)
"The benchmark", measurement_container_benchmark.cpp is in the repo.

I will fix the comment, does that clear things up?

navganti:
@leokoppel okay that makes sense, thanks.

cmake/WaveUtils.cmake
```diff
@@ -40,6 +40,31 @@ FUNCTION(WAVE_ADD_TEST NAME)
 
 ENDFUNCTION(WAVE_ADD_TEST)
 
+# wave_add_benchmark: Add a target which links against google benchmark
+#
+# WAVE_ADD_BENCHMARK(Name src1 [src2...])
```
navganti: @leokoppel, Do you mind renaming this file? We briefly talked about this on PR #226.

navganti: Okay I see #233. Will approve this then.

PR #229, opened 2017-10-26
Improve MeasurementContainer::get() performance

### LINKED ISSUE #227: Measurement container performance scales poorly

leokoppel:
Part of #227

The benchmarks added in #228 show `emplace()` performance is as expected, but `get()` is much slower than expected: O(NlogN) instead of O(logN).

Refactor the MeasurementContainer internals to remove the problem. As a bonus the new implementation is a bit simpler.

Details:
> Change the internal indexing of the multi_index_container: add a sensor-first
    composite index.
    Use this sensor-first composite index in get() instead of creating a view,
    which was really inefficient.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings

### DIFF
--- a/wave_containers/include/wave/containers/impl/measurement_container.hpp
+++ b/wave_containers/include/wave/containers/impl/measurement_container.hpp
@@ -139,7 +139,9 @@ typename MeasurementContainer<T>::ValueType MeasurementContainer<T>::get(
 
     if (i_next == sensor_index.end() || i_next->sensor_id != s) {
         // Requested time is not between two measurements for this sensor
-        throw std::out_of_range("MeasurementContainer::get");
+        throw std::out_of_range{
+          "MeasurementContainer::get: "
+          "requested time is after last measurement for sensor"};
     }
 
     if (t == i_next->time_point) {
@@ -150,7 +152,9 @@ typename MeasurementContainer<T>::ValueType MeasurementContainer<T>::get(
     // If no exact match, need at least one previous measurement to interpolate
     if (i_next == sensor_index.begin() || std::prev(i_next)->sensor_id != s) {
         // Requested time is not between two measurements for this sensor
-        throw std::out_of_range("MeasurementContainer::get");
+        throw std::out_of_range{
+          "MeasurementContainer::get: "
+          "requested time is before first measurement for sensor"};
     }
 
     // Requested time is between two applicable measurements. Can interpolate.
--- a/wave_containers/include/wave/containers/impl/measurement_container.hpp
+++ b/wave_containers/include/wave/containers/impl/measurement_container.hpp
@@ -148,16 +148,14 @@ typename MeasurementContainer<T>::ValueType MeasurementContainer<T>::get(
     }
 
     // If no exact match, need at least one previous measurement to interpolate
-    const auto i_prev = std::prev(i_next);
-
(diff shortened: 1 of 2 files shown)


wave_containers/include/wave/containers/impl/measurement_container.hpp
```diff
@@ -138,48 +129,35 @@ typename MeasurementContainer<T>::iterator MeasurementContainer<T>::erase(
...
+    if (t == i_next->time_point) {
         // Requested time exactly matches
-        return p_next->value;
+        return i_next->value;
     }
 
-    // If no exact match, need at least one more measurement to interpolate
-    if (i_next == time_view.begin()) {
+    // If no exact match, need at least one previous measurement to interpolate
+    const auto i_prev = std::prev(i_next);
```
navganti: Will this work? If `i_next` is already `sensor_index.begin()`, what will happen when you try to do `std::prev`?

wave_containers/include/wave/containers/impl/measurement_container.hpp
```diff
@@ -138,48 +129,35 @@ typename MeasurementContainer<T>::iterator MeasurementContainer<T>::erase(
...
+
+    if (i_next == sensor_index.begin() || i_prev->sensor_id != s) {
```
navganti: Also does this still hold? Because now the sensor could very well be the only sensor in the `sensor_index`?

navganti: Maybe I don't understand what the sensor view is. It's the list of all measurements for the sensor `s`, correct? Would there be a chance that `i_prev` has a sensor id that isn't `s`?

wave_containers/include/wave/containers/impl/measurement_container.hpp
leokoppel: Great catch Nav. It is undefined behaviour. I will change it

wave_containers/include/wave/containers/impl/measurement_container.hpp
leokoppel: There is no sensor view now. sensor_index has all the same measurements, just sorted by sensor first. Yes there is a chance.

wave_containers/include/wave/containers/impl/measurement_container.hpp
```diff
@@ -138,48 +129,35 @@ typename MeasurementContainer<T>::iterator MeasurementContainer<T>::erase(
...
 
-    // find the first element with time >= t
-    auto i_next = time_view.lower_bound(t);
+    // find the first element with sensor>=s and time >= t
+    auto i_next = sensor_index.lower_bound(boost::make_tuple(s, t));
 
-    if (i_next == time_view.end()) {
+    if (i_next == sensor_index.end() || i_next->sensor_id != s) {
         // Requested time is not between two measurements for this sensor
         throw std::out_of_range("MeasurementContainer::get");
```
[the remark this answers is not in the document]
leokoppel: Sure

PR #231, opened 2017-10-29
CMake: allow partial build and finding components

leokoppel:
This is a higher priority than I thought since we have some projects which need parts of libwave, but don't want to install all the dependencies.

- There is a BUILD_wave_<module> option for each module, which the user can set OFF to disable that module
- Each module checks its dependencies and disables its own build if any are missing
- There is a new helper, `WAVE_ADD_MODULE`, which does almost everything
- The user can call `find_package(wave COMPONENTS geometry ...)`
- Each module's dependencies are checked again, in case a dependency present at build time is no longer found.

I used "component" and "module" somewhat interchangeably, any preferences?

This dependency system is less sophisticated than those used by some large projects. For example, a module can't be disabled by default, the modules must be listed in top-level CMakeLists in a particular order with no circular dependencies allowed, and the messages could be prettier. However it is simple and enough for now.


CMakeLists.txt
```diff
@@ -21,19 +30,19 @@ OPTION(EXPORT_BUILD
...
  Allows the package to be found without install." OFF)
 
-# Find common dependencies (used by multiple required modules)
+# Find all dependencies here, and ensure they have IMPORTED targets
 
 # Require Eigen 3.2.92, also called 3.3 beta-1, since it's in xenial
-FIND_PACKAGE(Eigen3 3.2.92 REQUIRED)
-FIND_PACKAGE(Boost 1.54.0 REQUIRED system filesystem)
-FIND_PACKAGE(PCL 1.8 REQUIRED
+FIND_PACKAGE(Eigen3 3.2.92)
```
navganti: So now that it's not required...what will happen? I guess the module you're trying to include will require it, and it'll fail there?

README.md
```diff
@@ -64,15 +64,10 @@ your project's `CMakeLists.txt` file as follows:
     cmake_minimum_required(VERSION 3.0)
     project(example)
 
-    find_package(wave REQUIRED)
+    find_package(wave REQUIRED geometry matching)
```
navganti: :+1: cool

README.md
```diff
@@ -64,15 +64,10 @@ your project's `CMakeLists.txt` file as follows:
...
 
-    # Add a target depending on all of libwave
     add_executable(example example.cpp)
-    target_link_libraries(example wave::wave)
-
-    # Add a target depending on certain libwave components
-    add_executable(example2 example2.cpp)
-    target_link_libraries(example2 wave::geometry wave::containers)
+    target_link_libraries(example wave::geometry wave::matching)
```
navganti: wave::wave will still work if you want the whole thing, right?

wave_optimization/CMakeLists.txt
```diff
@@ -1,28 +1,20 @@
...
+WAVE_ADD_MODULE(${PROJECT_NAME} DEPENDS
     wave::utils
     wave::kinematics
     wave::vision
     Eigen3::Eigen
-    ceres)
+    ceres
+    SOURCES
+    src/ceres/ba.cpp
+    src/ceres/ceres_examples.cpp)
```
navganti: Just curious, do we need these ceres examples in libwave? Like I'm pretty sure they're the examples straight from their website?

README.md
leokoppel: Yes. But it's not a good idea to use it, since now you don't know what components are included. Maybe it should be removed?

wave_optimization/CMakeLists.txt
leokoppel: I thought they were original examples, but yes they don't add much to libwave..

CMakeLists.txt
leokoppel: `WAVE_CHECK_MODULE` will cancel build of the dependent module.

Example output:
```
-- Building wave_utils
-- Building wave_geometry
-- Not building wave_containers: Requires MissingDep
-- Not building wave_benchmark: Requires wave::containers
-- Building wave_controls
-- Building wave_kinematics
-- Building wave_matching
-- Not building wave_vision: Requires wave::containers
-- Not building wave_optimization: Requires wave::vision
```

README.md
leokoppel: Actually if you ask for REQUIRED components inf find_package, then yes you could link to `wave::wave` and be sure those components are included. Better to be specific anyway IMO, so you're not accidentally relying on something else.

CMakeLists.txt
leokoppel: Maybe I should make Eigen required anyway; it would be funny for the configure to succeed even though zero modules can be built.

README.md
navganti: Okay that makes sense.

wave_optimization/CMakeLists.txt
navganti: they're also the culprit for that whole Eigen/Ceres versioning debacle 😉

leokoppel: Yes well we will _actually_ use them soon... I hope

PR #232, opened 2017-10-29
Add package.xml

leokoppel:
Let libwave be built as part of a catkin workspace.

Note this does *not* make libwave a ROS package. It just lets catkin identify it as a pure CMake package to be built. Some packages we use, such as kindr and gtsam, do the same thing.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,15 @@
+<package format="2">
+  <name>libwave</name>
+  <version>0.1.0</version>
+  <description>libwave</description>
+
+  <maintainer email="chutsu@gmail.com">Chris Choi</maintainer>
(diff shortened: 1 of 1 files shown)


leokoppel:
```bash
cd `mktemp -d`
mkdir src
ln -s ~/code/libwave src/libwave  # replace with your own path to libwave
catkin build
ls devel/lib
```

Shows libwave is built and installed into the devel space.

leokoppel:
You can make a blank catkin package in the same workspace which does `find_package(wave REQUIRED)` and it finds the libwave from the workspace.

I think there are some additional issues e.g. with find_package(Eigen3) on xenial, to be fixed though.

PR #235, opened 2017-10-31
Add BUILD_TESTING option, tests target

leokoppel:
- Add `BUILD_TESTING` option
  - ON by default
  - Test targets are only added if it is ON
  - Why this name (instead of `BUILD_TESTS`)? It seems to be a standard CMake name, e.g. [used by CTest](https://cmake.org/cmake/help/v3.1/module/CTest.html)
- Add `tests` target
  - `make tests` builds the tests but does not run them; `make test` runs them.
  - Why this name in particular? Because catkin uses it


navganti:
why does `make test` run them? That's a little confusing compared to `make_tests`, no?

leokoppel:
`make test` is [automatically added by CMake](https://cmake.org/Wiki/CMake/Testing_With_CTest) and it's always been there. I was just mentioning it.

navganti:
Oops I meant `make tests`

leokoppel:
Yes it is confusing. I think the blame is on CMake for having "make test" not make anything.
See https://stackoverflow.com/questions/733475/cmake-ctest-make-test-doesnt-build-tests

Issue #234, opened 2017-10-30
Reduce wave_utils dependencies


leokoppel:
#231 allows partial build without all the dependencies, but wave_utils is still needed by most modules and requires opencv.

#### Current behaviour
- wave_utils depends on opencv, just to allow reading into cv::Mat from a yaml file.
- wave_utils contains wave_test.hpp, used by libwave unit tests

#### Expected behaviour
- wave_utils should not depend on opencv
- wave_test.hpp should not be in wave_utils, as it is not part of public interface

PR #236, opened 2017-11-01
Refactor ConfigParser

leokoppel:
Prereq for #234

Refactor ConfigParser. The primary motivation is to allow defining parsing functions in other modules, so wave_utils does not depend on OpenCV, or any future classes we want to parse.

Basically this is a wrapper around yaml-cpp's `convert` mechanism. However, by request I kept the interface to ConfigParser the same, so other classes still use `addParam`, `load(file)`, etc.

Note that despite the templates, the parsing functions for Matrix, etc. are still compiled into the library. Only the same types are still supported.

- Remove huge enum and switch statements of types
- Use templated Param class
- Define enum for return codes instead of int

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


wave_utils/include/wave/utils/config.hpp
```diff
@@ -131,93 +128,125 @@ class ConfigParser {
...
+ *   cols: 3
+ *   data: [1.1, 2.2, 3.3,
+ *          4.4, 5.5, 6.6,
+ *          7.7, 8.8, 9.9]
+ * ```
+ *
+
+ */
+template <>
+struct convert<cv::Mat> {
```
chutsu: Was this not supposed to be in the vision module?

wave_utils/include/wave/utils/config.hpp
```diff
@@ -25,57 +25,54 @@ namespace wave {
...
+    KeyError = -2,       ///< key not found, parameter is not optional
+    ConversionError = -3 /**< Invalid value for conversion
+                           * e.g. wrong-size list for vector, missing
+                           * key 'rows', 'cols' or 'data' (for matrix) */
+};
+
+/** Base class representing a parameter to be parsed in the yaml file */
+struct ConfigParamBase {
+    ConfigParamBase(std::string key, bool optional)
+        : key{std::move(key)}, optional{optional} {};
```
chutsu: why move? typo?

wave_utils/include/wave/utils/config.hpp
```diff
@@ -131,93 +128,125 @@ class ConfigParser {
...
+    ConfigStatus getYamlNode(std::string key, YAML::Node &node);
+
+    /** Check whether a key is present in the yaml file */
+    ConfigStatus checkKey(std::string key, bool optional);
+
+    /** Load yaml param primitive, array, vector or matrix. */
+    ConfigStatus loadParam(const ConfigParamBase &param);
+
+    /** Load yaml file at `config_file`. */
+    ConfigStatus load(std::string config_file);
```
chutsu: `const std::string &` maybe, I know I haven't been putting `const` everywhere but maybe we should?

wave_utils/include/wave/utils/config.hpp
leokoppel: See https://stackoverflow.com/questions/7592630/is-pass-by-value-a-reasonable-default-in-c11
https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html

wave_utils/include/wave/utils/config.hpp
leokoppel: Separate PR #237

wave_utils/include/wave/utils/config.hpp
leokoppel: will do

wave_utils/include/wave/utils/config.hpp
chutsu: This just hurts my head.

leokoppel:
OK, I added some reasoning to the issue. One of our projects does currently have an issue with opencv  which I can tell you about offline.

PR #237, opened 2017-11-01
Make wave_utils not depend on opencv

leokoppel:
Part of #234 

- Remove wave_utils dependency on OpenCV
- Move the cv::Mat parser to wave_vision 
- (tacking on) request correct Ceres version
    * We use EigenQuaternionParameterization somewhere. Currently will not build without 1.12

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


navganti:
Ceres 1.12? But your ppa has 1.13?

leokoppel:
It is based on what is used in the code

> We use EigenQuaternionParameterization somewhere. Currently will not build without 1.12

PR #240, opened 2017-11-01
Fix headers install path

### LINKED ISSUE #239: Headers install to wrong path
#### Platform
xenial

#### Steps to Reproduce
1. `cmake .. -DCMAKE_INSTALL_PREFIX=/tmp`
2. `make install`

#### Expected result
Headers install to `/tmp/include`

#### Observed result
Headers install to `/tmp/include/include`. For example
`/tmp/include/include/wave/kinematics/pose.hpp`

leokoppel:
Resolves #239 

Test:
```
cmake .. -DCMAKE_INSTALL_PREFIX=/tmp
make install
```

Use CMake install correctly:
> The last component of each directory name is appended to the destination directory but a trailing slash may be used to avoid this because it leaves the last component empty.

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


navganti:
labelling as hot-fix, because @Jebediah and I are very sad over here

navganti:
Thank you for fixing this so quickly @leokoppel

PR #249, opened 2017-11-16
Add explicit dependency on opencv modules

leokoppel:
Resolves reported build errors.
The vision module depends on ~xfeatures2d~ videoio, but did not specify the dependency.

**Previous behaviour**
When videoio module is not installed, wave_vision is still configured to build and fails during build.

**New behaviour**
When videoio is not found, wave_vision is disabled and not built (this prints a status message, not a warning or error).

### DIFF
--- a/wave_vision/CMakeLists.txt
+++ b/wave_vision/CMakeLists.txt
@@ -8,6 +8,7 @@ WAVE_ADD_MODULE(${PROJECT_NAME}
     Eigen3::Eigen
     Boost::filesystem
     opencv_core opencv_features2d opencv_calib3d opencv_imgproc opencv_highgui
+    opencv_videoio
     SOURCES
     src/utils.cpp
     src/dataset/VoDataset.cpp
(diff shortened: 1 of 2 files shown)


wave_vision/CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ WAVE_ADD_MODULE(${PROJECT_NAME}
     Eigen3::Eigen
     Boost::filesystem
     opencv_core opencv_features2d opencv_calib3d opencv_imgproc opencv_highgui
+    opencv_videoio opencv_xfeatures2d
```
navganti: I'm not using xfeatures. Anything from the contrib modules I don't think we can use since libwave is open source...our licensing would have to change.

navganti: If you could just remove it for now that would be great :)

navganti: WAIT I just read the message. What depends on xfeatures?!

PR #250, opened 2017-11-17
Changes for building against the new qnx_stage

mantkiew:
Changes according to https://bitbucket.org/renesas-astbd/renesas-demo/pull-requests/2/changes-for-building-against-qnx_stage/diff


mantkiew:
Ah, in that case, we just need to update the libwave submodule in autonomoose.

Issue #254, opened 2017-11-28
GroundSegmentation unexpected behaviour


leokoppel:
#### Platform
xenial

#### Steps to Reproduce
1. Run wave_matching_ground_segmentation_tests 
2. During the test the clouds are visualized in this order: "ground", then "drv", then "obs"

#### Expected result
The ground appears, then overhanging obstacles that are "drivable" under (presumably.. right?), then obstacles that are not drivable.

#### Observed result
Ground shows up fine, followed by the walls , followed by the bottom section of the walls.

PR #256, opened 2017-11-29
Make GroundSegmentation a templated filter

### LINKED ISSUE #253: Improve GroundSegmentation interface
#### Current behaviour
- GroundSegmentation only works on clouds of `PointXYZ`
- Unintuitive seqence of constructor, setup(), segmentGround(), and copyPointCloud() calls are needed to use

#### Suggested changes
- Make GroundSegmentation a pcl Filter as described here: http://pointclouds.org/documentation/tutorials/writing_new_classes.php
- Remove PointXYZGD struct

leokoppel:
Resolves #253 

 - Inherit `pcl::Filter`
 - Template on point type
 - Instead of copying points, work with indices
 - Remove PointXYZGD
 - Add setKeepGround(), setKeepObstacle() and setKeepOverhanging() methods for choosing which parts are wanted. By default just ground is removed.

This is a refactoring and interface change only. The previous buggy behaviour (#254) is kept.

Testing:
`wave_matching_ground_segmentation_tests` output is the same (compared screenshots before and after)

#### Pre-Merge Checklist:
- [ ] Code has ~automated~ visual tests
- [x] Zero compiler warnings


wave_matching/include/wave/matching/ground_segmentation.hpp
```diff
@@ -41,16 +42,14 @@ struct SignalPoint {
...
-    std::vector<PointXYZGD, Eigen::aligned_allocator<PointXYZGD>>
-      obs_points;  // just the obs points
-    std::vector<PointXYZGD, Eigen::aligned_allocator<PointXYZGD>> drv_points;
-    std::vector<PointXYZGD, Eigen::aligned_allocator<PointXYZGD>>
-      ground_points;  // just the ground points
-    PointXYZGD prototype_point;
-    int cluster_assigned;  // what cluster is it assigned to
-    Vec3 obs_mean;         // mean of obstacle points
+    // Points in the input cloud are referred to by indices
+    std::vector<int> bin_indices;  // all the points
```
Jebediah: Would it make more sensor for the type to be unsigned?

wave_matching/include/wave/matching/ground_segmentation_params.hpp
```diff
@@ -24,6 +24,7 @@ struct GroundSegmentationParams {
     // considered ground
     float p_tg = 0.3;  // ground height threshold
 
+    /// Height the robot (m), used to distinguish "drivable" overhanging points
     double robot_height = 1.2;
```
Jebediah: So that's what this does!

wave_matching/include/wave/matching/ground_segmentation.hpp
leokoppel: PCL functions take `std::vector<int>`, that's why I used this

PR #257, opened 2017-11-29
Fix ground segmentation bug

### LINKED ISSUE #254: GroundSegmentation unexpected behaviour

leokoppel:
Fixes #254

Before: both low and high points labeled as obstacles, low obstacles labeled as overhanging ("drivable")
Now: low points labeled obstacles, high points labeled overhanging

#### Pre-Merge Checklist:
- [ ] Code has automated tests - visual test only
- [x] Zero compiler warnings

### DIFF
--- a/wave_matching/include/wave/matching/impl/ground_segmentation.hpp
+++ b/wave_matching/include/wave/matching/impl/ground_segmentation.hpp
@@ -309,19 +309,19 @@ void GroundSegmentation<PointT>::sectorINSAC(int sector_index) {
             } else {
                 // check drivability
                 if (h > this->params.robot_height) {
-                    // drivable
-                } else {
-                    this->drv_indices.push_back(j);  // add to obs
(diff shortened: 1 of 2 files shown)


Jebediah:
Do you have a test scan with overhanging obstacles?

leokoppel:
No. From looking at the paper, the top of a wall is considered "overhanging", so is there anything special about actually overhanging objects? How could we tell if it's working?  @arunxdas

PR #265, opened 2017-12-12
Fix match() selection, censi

leokoppel:
- Fix switch statement
- Disable a test, todo @Jebediah 
- Simplify Censi inverse. It does change output (making covariance diagonal centimetres instead of  hundreds of metres). I am not sure what the self-adjoint was meant to do before. Another todo for tests.

#### Pre-Merge Checklist:
- [ ] Code has automated tests
- [x] Zero compiler warnings

### DIFF
--- a/wave_matching/src/icp.cpp
+++ b/wave_matching/src/icp.cpp
@@ -134,9 +134,13 @@ bool ICPMatcher::match() {
 
 void ICPMatcher::estimateInfo() {
     switch (this->params.covar_estimator) {
-        case ICPMatcherParams::covar_method::LUM: this->estimateLUM();
-        case ICPMatcherParams::covar_method::CENSI: this->estimateCensi();
-        case ICPMatcherParams::covar_method::LUMold: this->estimateLUMold();
+        case ICPMatcherParams::covar_method::LUM: this->estimateLUM(); break;
+        case ICPMatcherParams::covar_method::CENSI:
+            this->estimateCensi();
+            break;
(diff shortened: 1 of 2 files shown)


Jebediah:
The self adjoint thing is a way to specify that the matrix is symmetric

leokoppel:
Does my comment not explain clearly, or is it wrong?

```
        // The covariance is approximately: (Prakhya eqn. 3)
        // d2J_dX2^-1 * d2J_dZdX*cov(z)*d2J_dZdX' *  d2J_dX2^-1
        // = d2J_dX2^-1 * middle * d2J_dX2^-1
        //
        // To get the information matrix, we apply the inverse
```

cov(X) = d2J_dX2^-1 * middle * d2J_dX2^-1
cov(X)^-1 = (d2J_dX2^-1 * middle * d2J_dX2^-1)^-1
cov(X)^-1 = d2J_dX2 * middle^-1 * d2J_dX2

leokoppel:
Oh I see, I did not notice that `d2J_dX2` only contains the top half of d2J/dX2, and selfAdjointView is needed to fill the other half.

Jebediah:
Oh I see. I assumed that because the output was different you changed something. How is the output different now?

Jebediah:
What happens if you put the self adjoint back in but keep the change to the inverse? Does it produce the original output again?

leokoppel:
Yes, it produces the same output (other than numerical error). Will close for now

PR #274, opened 2018-02-20
CMake: set compatibility to ExactVersion

leokoppel:
We are breaking API all the time due to development.

- Add a note about API breaking to ReadMe
- Set CMake compatibility to `ExactVersion`, meaning a dependee asking for libwave 0.1.0 will get an error if 0.1.1 or 0.2.0 is installed.

Eventually we want to move to [semantic versioning](https://semver.org/), but currently we are in what that standard calls the "initial development phase":

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

We should start incrementening the minor version though.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,9 +131,12 @@ CONFIGURE_PACKAGE_CONFIG_FILE(
     INSTALL_DESTINATION "${INSTALL_CMAKE_DIR}")
 
 # Generate the Version file
+# This build is only considered compatible with the exact version requested
+# (not SameMajorVersion) for now since the API is often broken by active
+# development.
 WRITE_BASIC_PACKAGE_VERSION_FILE(waveConfigVersion.cmake
     VERSION ${WAVE_PACKAGE_VERSION}
-    COMPATIBILITY SameMajorVersion)
+    COMPATIBILITY ExactVersion)
 
 # Install the Config and ConfigVersion files
 INSTALL(FILES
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -131,9 +131,12 @@ CONFIGURE_PACKAGE_CONFIG_FILE(
     INSTALL_DESTINATION "${INSTALL_CMAKE_DIR}")
 
 # Generate the Version file
+# This build is only considered compatible with the exact version requested
+# (not SameMajorVersion) for now since the API is often broken by active
+# development.
 WRITE_BASIC_PACKAGE_VERSION_FILE(waveConfigVersion.cmake
     VERSION ${WAVE_PACKAGE_VERSION}
-    COMPATIBILITY SameMajorVersion)
+    COMPATIBILITY ExactVersion)
```
navganti: Sorry, what all does this entail? So for any dependencies they must have the same exact version as opposed to the SameMajorVersion?

leokoppel:
Yes. It does potentially add a lot of work: if someone in a large monorepo project were to update the libwave submodule, modules may refuse to build until their CMakeLists are explicitly updated with the new libwave version. But it is more honest about our (lack of) compatibility guarantee.

navganti:
I think that's fine for now @leokoppel. Plus this way the user will be more conscious about their libwave version and why they would like to use it. Better than everything breaking and the user getting mad at us :(.

PR #290, opened 2018-07-30
Updates to install scripts

Kaythay:
I took the commits from the [update_install_scripts](https://github.com/wavelab/libwave/tree/update_install_scripts) branch that were related to the install scripts, and cherry-picked into this branch. The original commits are:
https://github.com/wavelab/libwave/commit/fa90b3828ce1cea895ece4965c1a2f92af2dcd7d
https://github.com/wavelab/libwave/commit/bd6a202f8a4209c472d3f89cde0eb57b8773a55f
https://github.com/wavelab/libwave/commit/2f27e4ed63b36178244f71b9ea364c21fdca5cbb
https://github.com/wavelab/libwave/commit/1bc4ec2572b0534f610b3fbf464396cb7a3fd0ec
https://github.com/wavelab/libwave/commit/cf0842fba8b7955288d7baededc0ead935b4c077

#### Pre-Merge Checklist:
- [x] Zero compiler warnings


CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ OPTION(BUILD_BENCHMARKS
 
 
 # Find all dependencies here, and ensure they have IMPORTED targets
-
+LIST(APPEND CMAKE_PREFIX_PATH "/opt/wavelab")
```
leokoppel: Can you make this an [option](https://cmake.org/cmake/help/v3.5/command/option.html)? The user should be able to disable adding this path.

You can also add a status message like "Adding /opt/wavelab to CMAKE_PREFIX_PATH", and a comment here that some of our  install scripts put stuff here. (it's not "all dependencies"). Just to avoid [surprising the user](https://en.wikipedia.org/wiki/Principle_of_least_astonishment)

Kaythay:
I'm going add a commit that updates the README, so please don't merge the PR yet.

README.md
```diff
@@ -14,17 +14,40 @@ This library contains reusable code for:
...
 - Ceres 1.13
 - Eigen 3.3.2
-- Kindr 1.0.4
 - OpenCV 3.2.0
 - PCL 1.8
 - yaml-cpp 0.5.1
 - CMake 2.8.3
 - GCC 5.4
+- gtsam
+- GeographicLib 1.49
```
Kaythay: Is this the right version number for GeographicLib?

leokoppel:
Hmm, can we remove the `install.bash` entirely, or at least make it not seem like an obvious one-size-fits-all solution in the readme?

I think it's purpose was to quickly install a bunch of dependencies on our dev machines (it doesn't  `install` libwave, actually). For the average user, this may be unexpected and break things -- _or_ now that we semi-hide the dependencies in `/opt/wavelab`, they may not expect that it only installs some arbitrary subset of dependencies in a weird path.

Since we have a modular build now (no errors when missing a single dep such as PCL; just a warning), I think the user experience of "Oh, `wave_matching` is not building because I need to install PCL" is fine.

The other steps done by the install script (`git clone --recursive`, `cmake ..`, `make install`) are standard and should be just fine to list in the readme. Thoughts @navganti @chutsu ?

navganti:
@leokoppel I agree. If we have an `install.bash`, that in theory should do EVERYTHING, sort of like the anm_install that we have.

As long as the proper installation instructions are clear in the README then I am good.

chutsu:
Just call it `install_deps.sh`?

leokoppel:
@chutsu we already have an `install_deps.bash`. `install.bash` additionally does `git clone` and `make`. So I'll take your comment as in favour of deleting `install.bash`, and keeping `install_deps.bash`.

